PR #123, opened 2021-01-29
In docs/manage/platform-mgmt added section `Add / Remove Collaborators`

jedbgold:
@taylordowns2000 I'm starting to get more comfortable with this. Please see my addition to 'Add Remove Collaborators Section'. Note that we should probably discuss screen shots and aesthetic principles to keep in mind.


taylordowns2000 (changes requested):
Hey @jedbgold , I'm rejecting this because it's not passing the build tests. [See here](https://github.com/OpenFn/docs/pull/123/checks?check_run_id=1789296170#step:4:36) for what seems to be the problem. When you tested this out on your local machine before submitting, did it all look good?

Other notes:
1. Line 162 and 168: Where is the word colleague coming from? Do you mean admin?
2. What are the little parenthetical numbers at the right of lines 165, 169, 172, and 175?
3. Why is there an empty L2 "Grant support access" heading on line 181?

PR #130, opened 2021-02-03
Added "Deployment Option" Article and "Recover Password" section

jedbgold:


taylordowns2000 (approved):
reverted sidebar changes, made a few deployment article edits, added truncation to the articles.

all looks good except password, which i've opened a new issue for.

PR #133, opened 2021-02-05
Lost Password Section with Longer gif

jedbgold:


taylordowns2000 (changes requested):
looks like there's identifiable information in the gif. please re-record with a search filter in your inbox that's only showing messages from the openfn system

PR #137, opened 2021-02-09
added section to apps re tableau and added tableau as a sidebar item …

jedbgold:
Please note this section is still a work in progress, but it might be good to get something onto the website sooner rather than later. I'm not sure how I feel about the diagram, but I was just thinking to myself "If I am a non-technical dalberg consultant, what would I need to understand this...?"


jedbgold:
@taylordowns2000 this is the lucidchart where I created the diagram:

PR #138, opened 2021-02-09
Added Tableau section and Tableau sidebar item

jedbgold:
Put together a placeholder for Tableau based on our conversation today.


taylordowns2000 (approved):

taylordowns2000:
@jedbgold , note change to the tableau image. logos could be whatever, but the first image seemed to me like OpenFn was somehow sitting inbetween tableau and the tableau data source. and this is not the case.

PR #145, opened 2021-02-15
Designing integrations

jedbgold:
Created stub for designing integrations. Will continue adding to it throughout the week. Let me know if you think the current structure works. What I'd like to do is walk citizen integrators through the design process by providing a hypothetical scenario and then walking them through step by step how to think about the different layers involved in a flow. Defining the flow is critical before introducing any automation tool like OpenFn...


docs/getting-started/designing-integrations.md
```diff
@@ -3,12 +3,147 @@ title: Designing Integrations
...
+various user groups, those purpose are often related and we find ourselves in a
+position where we need to connect the data that's in those different tools we
+use.
+
+If we are administering surveys using ODK, we might need to send metadata to a
+management system like Salesforce where we can analyze it, report on it and
+share it with other members of our team—maybe tracking the progress of a large
+study.
+
+In connecting these different tools we set up different data integration
```
taylordowns2000, line 65: @jedbgold , i got here with my edits... would you move from this point forward checking each sentence for content-value? words cost $100, sentences cost $1000.

taylordowns2000 (changes requested):

PR #146, opened 2021-02-17
Added some more information regarding webhooks

jedbgold:
Source of information is from the API minibook. Not sure if build/generic_sources is the right location for this. Curious to hear your thoughts.


taylordowns2000 (approved):
This is a strange voice for docs, in my opinion @aleksa-krolls , but I think it's good content. Is this word-for-word from the minibook? And should we do anything with the minibook?

PR #183, opened 2021-06-02
added design section

jedbgold:
Ready for your review. Added some preliminary 'design' sections.


willip (commented):
@taylordowns2000 in light of today I think we should hold off on merging (both because of the many WiP sections that we can fill out before publishing and because I do think some of this would need redoing in light of that chat

PR #207, opened 2021-07-14
Kobo update

ritazagoni:


taylordowns2000 (approved):

aleksa-krolls:
@ritazagoni Before we merge, can you please...
1. Add some notes about Kobo `questions` vs. `labels` (for questions, as well as how this works for `choices`)
2. Guidance on how to download the Kobo form/ export the metadata
3. Links to ConSoSci training materials, other resources that might be helpful
4. Test locally to make sure the Markdown will look good on prod, nothing will break the docs site when merging

PR #209, opened 2021-08-03
Create ckan.md

aleksa-krolls:


taylordowns2000:
@hnyende , @aleksa-krolls , could you run the formatter for these files? Lots seem to go way beyond 80 characters per line and have some non-standard breaks. In your editor, type `command+shift-p` to open to the action menu and then type "format" to find the "Format Document" command. So long as you have "Prettier" installed in your VsCode instance, it should follow our official style guide, outlined in the `.prettierrc` file.

PR #224, opened 2021-09-16
Docs on switching adaptor version locally

aleksa-krolls:


taylordowns2000 (approved):
thanks @aleksa-krolls !

PR #237, opened 2021-10-30
javascript learnings part 1

daissatou2:
Completed the first 4 modules. There are 4 remaining and I expect to finish the course next week. We can decide if I should add on to this article or create a new one :)

### DIFF
--- a/articles/2021-10-29-how-learning-javascript-helped-me-better-understand-jobs.md
+++ b/articles/2021-10-29-how-learning-javascript-helped-me-better-understand-jobs.md
@@ -1,0 +1,106 @@
+---
+layout: post
+title: How learning JavaScript is helping me better understand OpenFn jobs
+author: Aicha Diallo
+author_url: https://github.com/daissatou2
+author_image_url: https://avatars.githubusercontent.com/daissatou2
+tags: [javascript, tips, jobs]
+featured: true
+---
+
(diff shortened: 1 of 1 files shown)


aleksa-krolls:
@daissatou2 I learned something in this one, and would be excited to see a part two :-) 

@taylordowns2000 can you review for accuracy? 
Also, wondering if there are any other resources we should be linking to?

taylordowns2000 (changes requested):
Looks great, thanks @daissatou2 . Please note a few modifications I've made, and re-work the falsy values section:
1. Remember to use `<!--truncate-->` so that your _whole_ article doesn't show up in the list view
2. Keep contrasting examples to a _SINGLE_ contrast! (Think of the scientific method :-) ). If you're contrasting two ways of declaring a function and you want to highlight `=>`, make the use of `=>` the _only_ difference. This will focus the reader on your point—not leave them staring at the 12 other differences between the examples wondering what's most important.
3. See my caveat to your `=>` definition.
4. I think the "Falsy Values" section needs re-working. "For instance, if a student previously lived in Ratoma and then moved to an unknown district marked as "Other", "undefined" would not update the student's district in Salesforce but the empty string would." This sentence makes me question the whole section, because `""` _is_ a falsy value. Can you rework this bit?

daissatou2:
@taylordowns2000 addressed your feedback and added more content here from the modules I worked on on Friday.

aleksa-krolls:
@daissatou2 looks like changes have been requested...?  Can you pls review, resolve, and comment at Taylor when addressed and you think this is ready to be merged?

daissatou2:
@aleksa-krolls this is ready for you. The changes were added already.

PR #250, opened 2021-12-04
Improve Credentials Page

elias-ba:


taylordowns2000 (changes requested):
Looks like a bunch of the existing documentation has been lost in this PR. What's the rationale? Stuff like creds only being owned and editable by a single user feels pretty fundamental.

I think it would be worth describing each type of credential for the generic ones, and to follow the same grouping as we do on the platform. For the specific ones I think you're on the right track.
image

A final thought is that the language is pretty flowery. I think we should try to keep value words out of documentation as best we can. If I go to docs and see anything that sounds like "OpenFn is great" I'll leave right then and there. (Even if I do think OpenFn is great!)

PR #275, opened 2022-05-09
rewrite platform walk through

amberrignell:
Will add an [advanced] section to it later, to add a flow job and some simple transformation logic.


taylordowns2000 (changes requested):
Looks great @amberrignell . Please replace the `add image here` lines with something more descriptive—that text inside the square braces defines the `alt` tag for the images.

As a bonus, while you're in there, why not add an image title for the hover?

- from: `add image here`
- to: `Autoprocess`

taylordowns2000 (approved):

PR #282, opened 2022-07-01
commcare walkthrough

daissatou2:


aleksa-krolls:
@taylordowns2000 okay if we merge this? Will be something new to tweet about and use in new user registrations. @daissatou2 is working on a Kobo one as well :)

PR #295, opened 2022-09-28
Update docs for null state cron job

elias-ba:


taylordowns2000 (approved):
will merge when null fix is in

PR #328, opened 2023-02-16
remove `devtools` docs

mtuchi:
#### Implementation checklist
- [x] Ensure all devtools content is now handled in either CLI docs or in Adaptors docs
  - [x] there are 14 files with `devtools` mentions; each may be handled differently
  - [x] some documentation from the `devtools` page itself should be _moved_ to the new small page on `core` - but also marked as deprecated
- [x] Add a small section on running with `core` - add a notice that core is reaching EOL this year.
- [x] Remove all devtools content and references from docs (for the devtools page itself, mark as deprecated)
- [x] Ensure that any examples in docs-repo specific adaptor documentation make use of CLI, not devtools and/or core.


taylordowns2000 (approved):
made some small changes, but this looks good.

PR #335, opened 2023-03-16
updated the data model section

daissatou2:


aleksa-krolls (changes requested):
@daissatou2 some feedback to element
1. Pls specify that the entities you summarized are key entities for the `Aggregate` module of DHIS2, but see the `Tracker` docs for details on this data model. 
2. Can you please remove the extra bracket in the `App Overview` section? (this was my typo)
Screen Shot 2023-03-16 at 4 06 01 PM
3. Can you remove the entire `OpenFn Adaptor` section on this page? (Now that the docs are consolidated, ppl will get to this docs page via Docs/Adaptors... so this is redundant and has a link to the old adaptor repo.)
4. I don't think we should remove the link to the DHIS2 docs
```
See DHIS2 documentation for more about the data model and specific entities: https://docs.dhis2.org/archive/en/2.30/developer/html/techarch_data_model.html
```

aleksa-krolls (approved):

PR #346, opened 2023-05-09
Update `CLI` docs

### LINKED ISSUE #339: Add documentation on how to pass state from 1 job to another when working locally vs on platform
The OS4H team expressed confusion on how to pass state from one job to another when working locally during one of thier homework exercises. Please add documentation to docs.openfn.org which explains how to do this and how this is different from working on the platform. Consider creating a new docs page in the Build -> Jobs section, or expanding on this existing [page](https://docs.openfn.org/documentation/jobs/state) which discusses `state`.

mtuchi:
#### Description
Update CLI developer challenge to use workflow execution plan feature, Fixes #339

### DIFF
--- a/adaptors/salesforce.md
+++ b/adaptors/salesforce.md
@@ -146,6 +146,12 @@ values (e.g., `M: male, F: female`).
 semicolon-separated strings (e.g.,
 `Services__c: 'Food;Counselling;Medical_Aid`).
 
+#### Timeout considerations  
+
+The Salesforce adaptor currently waits for the batch to finish syncing to Salesforce to ensure that the batch results (i.e., whether the batch import was successful, how many records succeeded, how many failures) can be relayed back to the OpenFn user in Activity History. This can potentially cause timeout errors in OpenFn if the batch sync triggers Salesforce automation which takes long to complete. 
+
+In future adaptor updates, OpenFn will consider adding a new option to the adaptor to not wait for the batch results, and to have the batch run asynchronously. The tradeoff is that the OpenFn run would always succeed–and the administrator would have to log into Salesforce and monitor the batch results via the Salesforce Setup “Bulk Data Load Jobs” page. With this approach, OpenFn wouldn’t set limits on the batch processing time avoiding timeout errors, but it does move the batch monitoring from OpenFn to Salesforce. 
+
 ### Salesforce Credentials
 
 Salesforce requires a username, password, login URL, and security token to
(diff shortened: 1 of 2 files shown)


daissatou2:
This is a great addition to the CLI page! I know the OS4H team struggled with the concept of running flow jobs locally. Some feedback to consider: 

1. Please specify if job `id` is something that should be made up. Also, should it just be unique to that specific workflow plan file? 
2. Can you name the `id` for the starting job in your example workflow something other than "start"? That confused me a bit. 
3. Can we give an example of a realistic workflow plan? Maybe `getUsers` and then `postUsers`? **With the current docs, I don’t understand how I could format the execution plan so that the output of `getUsers` could automatically be used as the input of `postUsers`.**  Is it not possible to use a file name as the `data`? 
4. What is the difference between “execution plan”, “workflow plan” and workflow execution plan”? I’m seeing all three seemingly referring to the same thing. I think it would be a good idea to use one term. 
5. Can you add the text in italics to the docs? “Execution plan is a powerful feature of @openfn/cli that allows you to define a list of jobs and rules for executing them. You can use an Execution Plan to orchestrate the flow of data between systems, and to handle errors and retries in a structured and automated way. _For example, if you have two jobs in your workflow (GET users from system A & POST users to system B), you can use your execution plan to automatically run both jobs from start to finish. This imitates the [flow and fail trigger patterns](https://docs.openfn.org/documentation/build/triggers#flow-triggers) on the OpenFn platform where a second job should run after the first one succeeds or fails, respectively, using the data returned from the first job._ “
6. After the code snippet on executing the workflow plan (openfn workflow.json -i), include a sentence of what would happen upon execution. For instance “On execution, this workflow plan will first run x job using  x data, then run x job using x data”

taylordowns2000:
@mtuchi , please request a review from me when you've addressed @daissatou2 's changes! excited to get these docs up :-)

mtuchi:
@daissatou2 thank you for the feedback 👌🏽, i have address you comments and fine turned the language a bit,  Also i have added the examples to show how you can run the workflow. Please give it another spin and let me know if it's still confusing or if we can make it better

daissatou2:
@mtuchi I've added some time on your calendar for us to discuss my feedback.

daissatou2 (changes requested):
discussed feedback on a call with Mtuchi.

taylordowns2000 (approved):
approved from my end. when aicha approves the content we're good to go

daissatou2 (approved):

PR #358, opened 2023-08-14
update implementation checklist

daissatou2:


daissatou2:
synced with Mtuchi and the check failures are related to broken links on the adaptor docs. @mtuchi please let me know when these are resolved. Thanks!

mtuchi:
@daissatou2 i have fixed the docs build bug. It was caused by a broken process in adaptor build process

PR #394, opened 2024-02-09
docs sprint 2

daissatou2:


daissatou2:
@aleksa-krolls this PR includes 1) Roles page feedback, 2) BPMN diagramming Docs update & 3) History pages. See my notes on the History docs below:

1. I combined “Anatomy of the log view” & “Viewing Job Runs” into one video/page. I don't think we need a separate page for each of these.
2. I renamed “Search history & troubleshooting tips” to just Troubleshooting. I also borrowed heavily from the V1 troubleshooting [page]( https://docs.openfn.org/documentation/manage/troubleshooting-tips-on-platform). I updated the language to match V2 terms. 
3. We should add a video on bulk reprocessing, I don't see this in the infrastructure so I can work on this in the next sprint. 
4. For now I have linked to the 3 Loom videos. Would be nice if we could embed the videos on the pages and say “check out the video below”.

aleksa-krolls (approved):

Issue #393, opened 2024-02-09
CLI Challenge: COVID-19 stats endpoint not responsive


hunterachieng:
The COVID-19 challenge url is down and giving a `502` error status code when hitting the endpoint.

#### Potential Solution:
- Wait out and monitor the endpoint until it comes back online.
- Find another COVID-19 endpoint that provides similar data to be used.

#### Docs page to be updated
- https://docs.openfn.org/documentation/cli#-challenge-fetch-covid-19-metadata
- https://docs.openfn.org/documentation/next/cli-challenges#-retrieve-covid-19-metadata

aleksa-krolls:
@mtuchi any ideas here?

mtuchi:
Thanks @hunterachieng for flagging this, The second option seems viable `Find another COVID-19 data source`. All endpoint in https://disease.sh/ Seems to produce internal server error. 

From a quick google search i found this https://covidtracking.com/data/api. @hunterachieng, Can you spend 5 minutes to test that API, See if you can get metadata 

cc @aleksa-krolls

mtuchi:
Good news @hunterachieng , We have community contribution that address this issue, See #402 

There are just two tiny things that needs to be worked on,
- [ ] Fix broken Urls, [See the following failed checks](https://github.com/OpenFn/docs/pull/400/checks)
- [ ] Update platform CLI challenge https://docs.openfn.org/documentation/cli#-challenge-fetch-covid-19-metadata

cc @aleksa-krolls

aleksa-krolls:
@hunterachieng @mtuchi please note [my slack comment](https://openfn.slack.com/archives/C06HA4WF35K/p1707811642208039) re: how links to other should be updated (where you should reference file paths, and not page url paths)

PR #400, opened 2024-02-12
CLI Challenges amended with new COVID API and markdown formatting

bentleycook:
This changeset includes:

- Amended the COVID challenge to use a different API as the original appears to no longer be working.
- Markdown formatting likes for code blocks to include the language.
- Moved up the installation of http adaptor to the first challenge that expects a user to use it.

image

### DIFF
--- a/docs/build-for-developers/cli-challenges.md
+++ b/docs/build-for-developers/cli-challenges.md
@@ -4,7 +4,7 @@ sidebar_label: CLI Challenges
 slug: /cli-challenges
 ---
 
-#### Solve real-world problems and showcase your command-line skills by participating in our CLI challenges.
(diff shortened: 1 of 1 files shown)


bentleycook:
Some thoughts - the old COVID API keeps throwing 5XX and returning Cloudflare's gateway timeout view. I'm guessing it is gone for good.

The API I included in this PR is very US-centric, unfortunately. It isn't obvious to me what a compelling task for a user could be with this dataset, so I left it kind of open-ended.

mtuchi:
Thanks @bentleycook , We were discussing similar approach in here #393, I will review this and merge

mtuchi (approved):
These changes looks good, We had similar suggestion on #393

PR #445, opened 2024-03-20
Notes for travelogue

taylordowns2000:
@josephjclark , made a few notes [here](https://github.com/OpenFn/docs/pull/445/commits/88f5e41796253691404048499f0442b8c36ca671) - the first commit is just running the formatter so look at this one instead. call if you wanna chat about any of the proposed changes


josephjclark:
Thank you - I've pushed a counter commits!

We disagree I think on the grammar of `today-this very day-I will be better` vs `today - this very day - I will be better`. I think this is something about em-dashes? :shrug:  I like the spaced out version but I guess we do whatever is most consistent for the blog.

I'm really keen to keep the VAP stuff short, although I appreciate that I've been a little too flippant about it. I've taken another swing and pulled in some of your suggestions. I think it reads better now?

We just should just merge this into master and close the other branch. I also formatted it and I think the formatters may be clashing. I'm using prettier here which I thought  would just use the local `.prettierrc`?

taylordowns2000:
love it. good to go from my perspective.

josephjclark (approved):

josephjclark:
@taylordowns2000 I'll merge in the morning :)

josephjclark:
@taylordowns2000 turns out I don't have auth to merge to main. can you do the honours?

PR #467, opened 2024-04-09
Document step caching in the CLI

josephjclark:
Just some basic documentation to support https://github.com/OpenFn/kit/pull/645

### DIFF
--- a/docs/build-for-developers/cli-usage.md
+++ b/docs/build-for-developers/cli-usage.md
@@ -6,7 +6,8 @@ slug: /cli-usage
 
 This page shows common usage examples for the CLI.
 
-Execute a job, run a workflow, adjust logging, maintain adaptors, and save the state.
+Execute a job, run a workflow, adjust logging, maintain adaptors, and save the
+state.
 
 ---
 
@@ -54,9 +55,26 @@ openfn path/to/job.js -a http=/repo/openfn/adaptors/my-http-build
 openfn path/to/job.js -ma http
 ```
 
-Set a path to the monorepo with the env var OPENFN_REPO_DIR (eg, `OPENFN_REPO_DIR=~/openfn/repo openfn job.js -ma http`).
+Set a path to the monorepo with the env var OPENFN_REPO_DIR (eg,
+`OPENFN_REPO_DIR=~/openfn/repo openfn job.js -ma http`).
 
 Remember to rebuild the adaptor before using it!
+
+**Run from a specific start step**
+
+You can specify a step as an exact id, or a partial substring from the name or
+id.
+
+```bash
+openfn path/to.job.js --start cf628d9e -s path/to/input.json
+```
+
+If you have previously cached this workflow's results, the CLI will
+automatically load the cached input if you omit the `-s` argument:
+
+```bash
+openfn path/to.job.js --start cf628d9e
+```
 
 ---
 
@@ -76,6 +94,19 @@ openfn path/to/job.js -a adaptor-name -o path/to/output.json -s path/to/state.js
 ```bash
 openfn path/to/job.js -a adaptor-name -O
 ```
+
+**Save all step results to disk**
+
+```bash
+openfn path/to/workflow.json --cache-steps
+```
+
+Each step will write its output to `./cli-cache/<workflow-name>/<step-id>.json`.
+The `.cli-cache` folder will be git-ignored and the cache will be cleared when
+the workflow is re-run with `--cache-steps` enabled.
+
+To _always_ cache, set the `OPENFN_ALWAYS_CACHE_STEPS` env var to `"true"`, and
+pass `--no-cache-steps` to disable it.
 
 ---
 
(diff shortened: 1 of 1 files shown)


docs/build-for-developers/cli-usage.md
```diff
@@ -54,10 +55,39 @@ openfn path/to/job.js -a http=/repo/openfn/adaptors/my-http-build
 openfn path/to/job.js -ma http
 ```
 
-Set a path to the monorepo with the env var OPENFN_REPO_DIR (eg, `OPENFN_REPO_DIR=~/openfn/repo openfn job.js -ma http`).
+Set a path to the monorepo with the env var OPENFN_REPO_DIR (eg,
+`OPENFN_REPO_DIR=~/openfn/repo openfn job.js -ma http`).
```
mtuchi, line 59: @josephjclark i don't think i understand this part `OPENFN_REPO_DIR=~/openfn/repo openfn job.js -ma http`).` How does this work exactly ?

docs/build-for-developers/cli-usage.md
```diff
@@ -54,10 +55,39 @@ openfn path/to/job.js -a http=/repo/openfn/adaptors/my-http-build
...
+
+You can also pass `--end` to make the workflow exit early.
+
+**Run a single step**
+
+`--only` works just like `--start` and `--end`. You can partially match a step
+name or id, and input will be automatically loaded from the cache.
+
+```bash
+openfn path/to.job.js --only cf628d9e
```
mtuchi, line 88: I am just curious if i don't have cli cache, will this brow up `openfn path/to.job.js --only cf628d9e` i am curios if we need to emphasis on this will only work if you previous use `--cache-steps`

docs/build-for-developers/cli-usage.md
```diff
@@ -171,16 +214,16 @@ running workflows via the CLI.
 
 ### Load adaptor documentation
 
-The CLI can list adaptor documentation in the terminal. Note that it has to download
-the adaptor to the repo (if it's not already there), which can take a moment.
+The CLI can list adaptor documentation in the terminal. Note that it has to
+download the adaptor to the repo (if it's not already there), which can take a
+moment.
```
mtuchi, line 219: Just a question, After downloading the adaptor documentation 
Next time you run the `openfn docs http` for example. It will fetch from repo right ?

docs/build-for-developers/cli-usage.md
josephjclark: I'm not sure what's not to understand! :grimacing: 

This sets an env var (`OPENFN_REPO_DIR`) to a value before running the `openfn` command with a job, the http adaptor, and the monorepo flag.

Usually of course you'd set the env var in your `.zhrc` or whatever,

This diff is just caused by prettier - these aren't new docs or new functionality.

docs/build-for-developers/cli-usage.md
josephjclark: `--cache-steps` and `--only` are mutually exclusive and technically unrelated.

If you do `--only` without having cached steps:
- You can pass your own input with `-s`, which is likely what I'd expect you to do
- If you didn't pass state and the CLI couldn't find anything in the cache, you'll get a warning
- If the step doesn't need input anyway, it'll run just fine

docs/build-for-developers/cli-usage.md
josephjclark: From the local repo, yes. It will cache the documentation.

docs/build-for-developers/cli-usage.md
mtuchi: @josephjclark i am only familiar with these two 
```
export OPENFN_ADAPTORS_REPO=/Users/openfn/Workspace/adaptors
export OPENFN_ALWAYS_CACHE_STEPS=true
```

I have never used `OPENFN_REPO_DIR` before, that's why i am surprised

josephjclark: Oh! Right, that makes sense. Let me fix that - thanks!

josephjclark: Ach, sorry, you're right! Should be good now

mtuchi (approved):

docs/build-for-developers/cli-usage.md
```diff
@@ -48,16 +49,52 @@ openfn path/to/job.js -a http@2.0.0
...
 ```bash
 openfn path/to/job.js -ma http
 ```
 
-Set a path to the monorepo with the env var OPENFN_REPO_DIR (eg, `OPENFN_REPO_DIR=~/openfn/repo openfn job.js -ma http`).
+You must set the path to the monorepo in the env var OPENFN_ADAPTORS_REPO. For
+example:
+
+```bash
+OPENFN_ADAPTORS_REPO=~/openfn/adaptors openfn job.js -ma http
```
mtuchi, line 62: @josephjclark  this`OPENFN_ADAPTORS_REPO=~/openfn/adaptors openfn job.js -ma http` should be `OPENFN_ADAPTORS_REPO=~/openfn/adaptors`

josephjclark: No, it's correct as it is.

This runs a job with the monorepo path set in a var. It's an actually useful command.

If we just did `OPENFN_ADAPTORS_REPO=~/openfn/adaptors` then nothing would happen.

We could do `export OPENFN_ADAPTORS_REPO=~/openfn/adaptors` but that I think will only work within that session? So it's not terribly helpful.

This is why I added the line about adding to your `.profile`

mtuchi: Ooh okay, got you

PR #468, opened 2024-04-10
Self-guided migration steps

ritazagoni:
Instructions for how to partially automate the v1 to v2 migration steps


aleksa-krolls (approved):
Added changes

PR #469, opened 2024-04-11
Lazy State Docs

josephjclark:
This PR adds docs for the Lazy State operator.

I'm just going to call it the State Operator in public.

I am not ready to re-write all our docs examples using `$.data`, or remove the complicated "lazy state loading" section. I think it's appropriate for now to document this as an experimental new feature.

If it sticks (and I am sure it will), then we can overhaul the docs in a couple of months time to put the state operator front and center. @aleksa-krolls  asked for a single way to read from state and I think this is it.

This should be merged after the latest fix is released.

### DIFF
--- a/docs/jobs/job-writing-guide.md
+++ b/docs/jobs/job-writing-guide.md
@@ -3,8 +3,6 @@ sidebar_label: Job Writing Guide
 title: Job Writing Guide
 ---
 
-<!-- TODO: work out where this goes in the nav bar. Top level I think. I probably want to merge my structural changes first. -->
-
 Workflow automation and data integration in OpenFn is realised through the
 creation of Jobs.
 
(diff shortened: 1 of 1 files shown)


josephjclark:
Sketch of a post to Community:

Hi Community!

I'm really excited to share a shiny new thing with you: The State Operator!

In a nutshell, the State Operator `$` provides an easier, neater way to reference properties on state.

The magic is that you don't have to wrap your state paths in dataValue or arrow functions.

Here's how it works:
```
get($.data.url)
```
This is equivalent to doing this:
```
get((state) => state.data.url)
```
Or this:
```
get(dataValue('url'))
```

You can even use the lazy state operator in expressions - check out this template literal:

```
create({
  name: `${$.patients[0].first_name} ${$.patients[0].last_name}`
});
```

It is live on Lightning now (since vXX) and on the CLI since v (XX)

You can read more docs here.

We still consider the State Operator to be a bit experimental - so please help us out by experimenting! Try The State operator in your own code and let us know if it works well for you! Or, if it doesn't work, we'd love to hear that too.

Thanks!

aleksa-krolls (approved):

Issue #471, opened 2024-04-15
Display on required properties in sample configuration


mtuchi:
#### Description
Currently the `sample configuration` section on docs. Displays all configuration-schema properties regardless if they are required or not. This causes lot's of confusion especially for new users who are using the adaptor for the first time. 

### The request In few words
- [x] Display only required properties in sample-configuration schema
- [x] Make sure all properties type accurately matches the configuration schema type. For `Boolean` property it should show a `boolean` value and not a `string` value. Eg:[Postgresql adaptor ](https://docs.openfn.org/adaptors/packages/postgresql-configuration-schema). `ssl: true` not ~ssl: "true"~. Of

PR #492, opened 2024-05-17
Display only required properties in sample config

mtuchi:
#### Description
I have update the `sampleConfiguration()` function to only display required properties. 

Ref #471

### DIFF
--- a/generate-adaptors/index.js
+++ b/generate-adaptors/index.js
@@ -129,7 +129,7 @@ const sampleConfiguration = json => {
       conf[key] = 'Someting';
       conf[key] =
         Array.isArray(properties[key]['examples']) &&
-        `${properties[key]['examples'][0]}`;
+        properties[key]['examples'][0];
 
       return conf;
     });
--- a/generate-adaptors/index.js
+++ b/generate-adaptors/index.js
@@ -121,8 +121,15 @@ const sampleConfiguration = json => {
   let { properties, required } = json;
 
   const conf = {};
-  if (typeof required === 'undefined') {
-    return '```\n No required property \n```';
+  if (properties && typeof required === 'undefined') {
+    Object.keys(properties).forEach((key, index) => {
+      conf[key] = 'Someting';
+      conf[key] =
+        Array.isArray(properties[key]['examples']) &&
+        `${properties[key]['examples'][0]}`;
+
+      return conf;
+    });
   }
   if (properties && required) {
     required.forEach((key, index) => {
--- a/generate-adaptors/index.js
+++ b/generate-adaptors/index.js
@@ -123,7 +123,7 @@ const sampleConfiguration = json => {
   const conf = {};
   if (properties && typeof required === 'undefined') {
     Object.keys(properties).forEach((key, index) => {
-      conf[key] = 'Someting';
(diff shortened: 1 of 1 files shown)


josephjclark:
I didn't know this is how we did the sample config. I'm not a fan :grimacing: 

This actually gives us a worse appearance for http right? Because http has no required properties? So the schema will be empty.

I'd much rather have the option to manually define a sample schema somewhere and use that, rather than trying to autogenerate something.

mtuchi:
@josephjclark For adaptor that does not have a required property this is how the sample config will appear 

Screenshot 2024-05-18 at 11 19 18 AM

josephjclark:
Yeah that's my point - it's not very helpful!

"Required" and "example" are two very different things.

Ok, here's a compromise. If the schema has required schema, only show the required properties. Otherwise, show all properties with examples.

I still don't like it, and I'd rather put this time into a good lasting solution. But I'd probably accept it at least!

mtuchi:
I have thought of displaying all properties, but i find it misleading for http because it has both `username, password and token` See screenshot below 👇🏽 . Displaying all of these properties i think it's confusing especially for first time users. 

Should we utilize the `readme.md` For each adaptor and add the content for configuration ? or perhaps create a `configuration.md` file for each adaptor in adaptors monorepo ?
Screenshot 2024-05-20 at 9 27 17 AM

josephjclark:
An optional configuration.md in the adaptor repo would be the ideal. If there isn't one there, we default to showing all (or all required?) properties

mtuchi:
In our adaptor monorepo, currently we have tw `.md` files, `README.md` and `CHANGELOG.md`. We don'f have `configuration.md` file yet. Should i create an issue for adding `configugiration.md` in all adaptors, mean while we display all properties for adaptors that do not have required list ?

mtuchi:
Hiya @josephjclark , I have created an issue for creating the `configuration.md` file in each adaptor, See https://github.com/OpenFn/adaptors/issues/572 and also i have pushed the changes for displaying the full schema if `required` list is `undefined`

generate-adaptors/index.js
```diff
@@ -131,6 +131,16 @@ const sampleConfiguration = json => {
       return conf;
     });
   }
+  if (properties && required) {
+    required.forEach((key, index) => {
+      conf[key] = 'Someting';
```
josephjclark, line 136: Can we remove this line?

generate-adaptors/index.js
```diff
@@ -131,6 +131,16 @@ const sampleConfiguration = json => {
...
+      conf[key] =
```
josephjclark, line 137: Try
```
conf[key] = properties[key].examples?.[0]
```

What happens if there is no example and `conf[key]` is undefined?

mtuchi: I think if `conf[key]` is `undefined` that credential schema is not correct and should be investigated.

generate-adaptors/index.js
mtuchi: ✅ i have replaced that line with what you suggested 
```
conf[key] = properties[key].examples?.[0]
```

josephjclark (approved):

PR #497, opened 2024-05-26
docs on globals and job mapping tables

aleksa-krolls:
Adding guidance on how we use mapping tables in jobs and sometimes use tools like OCL


mtuchi (approved):
@aleksa-krolls This is good 👍🏽 ,I have pushed few typo fixes. 
You can proceed  with `Squash & merge`

PR #498, opened 2024-06-06
allow generate adaptors to run from local adaptors monorepo

josephjclark:
There's not really any way to preview changes made in the adaptors monorepo against the docsite. So you don't really know if its rendering right until you deploy.

To preview changes locally right now, you have to make a bunch of hacks in the code in the docsite AND the adaptors monorepo.

This PR makes it easy for the docs site to use the mono repo

## Usage

Set the `OPENFN_ADAPTORS_REPO` env var to point to the monorepo

Start the server with `yarn start:dev` to build from the monorepo on first load

If the server is running. run `yarn generate-adaptors -m` to rebuild from the adaptors list.

Note that the versions list is skipped if loading from the monorepo

## Docs

I've made a little guide in the adaptors repo https://github.com/OpenFn/adaptors/wiki/How-to-test-docs-changes

## TODO

* [x] Add some docs here
* [x] Should we use a flag as well as the env var? I think the env var is too squishy. But the solution has to work with build and docgen, so it can't just be a docgen CLI thing

### DIFF
--- a/generate-adaptors/index.js
+++ b/generate-adaptors/index.js
@@ -69,6 +69,8 @@ async function loadAdaptorsDocsFromMonorepo() {
   console.log('Loading adaptors docs from adaptors monorepo at ', baseDir);
 
   try {
+    // Read from tmp not from docs, because otherwise the adaptor
+    // build script won't work properly
     const raw = fs.readFileSync(path.resolve(baseDir, 'tmp/docs.json'));
     return JSON.parse(raw);
   } catch (e) {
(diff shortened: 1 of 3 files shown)


generate-adaptors/index.js
```diff
@@ -193,25 +215,51 @@ module.exports = function (context, { apiUrl }) {
...
+          }
+
+          const adaptors = await (useMonorepo
+            ? loadAdaptorsDocsFromMonorepo(process.env.OPENFN_ADAPTORS_REPO)
+            : loadAdaptorsDocsFromGithub());
 
           console.log('Generating adaptors docs via JSDoc...');
 
           adaptors.map(a => {
+            if (!a.name) {
```
josephjclark, line 258: Oh this is unrelated - I started getting weird errors locally for `cartodb`. I figure it's worth leaving this here just for better devx later

generate-adaptors/index.js
```diff
@@ -233,17 +281,16 @@ module.exports = function (context, { apiUrl }) {
               `./adaptors/packages/${a.name}-configuration-schema.md`,
               configurationSchemaBody
             );
-          });
-          console.log('Done ✓');
 
-          console.log('Creating sidebar paths...');
+            fs.writeFileSync(
```
josephjclark, line 285: I took the liberty of tweaking the output logs a bit :blush:

taylordowns2000 (approved):
fantastic. hey @mtuchi and @hunterachieng this will be a big help!

when making changes to your inline docs and examples for a new adaptor you can preview what they'll look like on the actual openfn docs site!!!

PR #500, opened 2024-06-17
collaboration page

ritazagoni:
New page to explain how to add/remove Project Collaborators


aleksa-krolls (approved):
Approved - but fyi @ritazagoni I noticed that some of the buttons in the app were named slightly differently than you had documented. But I went ahead and made the necessary updates

PR #507, opened 2024-06-21
Yarn v4.3.1

josephjclark:
Turns out yarn has its own self-versioning strategy now - it should really be locked to the project

I am not quite sure why, then, I was on v1. Maybe I'd missed some setup? I've just updated ubuntu, maybe that did something.

## QA notes

Can we run some quick QA on this?

- checkout this branch
- run `yarn install`
- should be no diffs in `yarn.lock`
- `yarn -v` should print 4.3.1?

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -38,6 +38,7 @@ jobs:
           USE_SSH: true
           GIT_USER: git
         run: |
+          corepack enable
           git config --global user.email "actions@gihub.com"
           git config --global user.name "gh-actions"
           yarn install
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -14,6 +14,8 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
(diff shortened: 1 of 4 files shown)


taylordowns2000 (changes requested):
ack!

image
image

josephjclark:
When I check out this branch my yarn version immediately changes to 4.3.1.

So I'm not sure why you're on 3.2.3 :thinking: 

Have you run `corepack enable` @taylordowns2000 ? Does that make any difference?

taylordowns2000:
just tried with `corepack enable` first and no luck. same diff (and same yarn version.)

i use asdf, would that impact?

josephjclark:
I'm using asdf too. But yeah it might be a factor. I don't know if an install or reshim makes a difference?

Frustrating issue. I don't care which version we use but I want all developers to be using the same yarn. Which should Just Work :tm: 

I'll run a couple of tests on a second system this afternoon

aleksa-krolls:
hey @josephjclark what's the status here?

josephjclark:
This and #506 are closely related

PR #510, opened 2024-06-25
Remove Aggregate, add Central

yanokwa:
We'll need to add to this over time, but it's an improvement over what exists now.

### DIFF
--- a/adaptors/odk.md
+++ b/adaptors/odk.md
@@ -2,7 +2,9 @@
 title: ODK
 ---
 
-In order to fetch data from ODK Central, you must run a scheduled job with a cron trigger the `odk` adaptor.
+In order to fetch data from ODK Central, you must run a scheduled job with a cron trigger using the `odk` adaptor.
+
+Here's an example of a job that gets submission metadata.
 
 ```js
 getSubmissions({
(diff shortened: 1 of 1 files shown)


josephjclark:
Hi @yanokwa ! Thank you so much for raising this. We didn't get a notification and I've just seen this now (the docs repo gets a bit less attention than some of our greater hits :sweat_smile: )

So just to be clear, it looks like we've got our adaptor wrong! Our `getSubmissions` is actually `getSubmissionsMetadata`.

I will raise a fix in the adaptor itself, make sure the API is all correct, then come back and take a look at these docs with you.

josephjclark:
Hi @yanokwa - we've made a fix on the adaptor along these lines. The PR is here: https://github.com/OpenFn/adaptors/pull/654  but it should be available early next week.

That means we need to work out what to do with the docs here! Because the examples you've given are out of date now - `getSubmissions()` should just work as you expect.

We'll take a look at the Overview next week and work out what to do. We'd love to  get your input on it if you have time though!

josephjclark:
I'm gonna merge this into a branch in this repo and re-open the PR. I'll fact check the contents so it's accurate with the latest changes, then catch up with @taylordowns2000 and @christad92 about the final copy. I know that there's some design work going on with Overview text - maybe this adaptor can set the standard

PR #511, opened 2024-06-27
Updated edge condition docs

### LINKED ISSUE #451: Conditional paths & JS expressions
We need more docs on how to write JS expressions for conditional paths, how to test & troubleshoot, and more examples. 

We have very little right now: https://docs.openfn.org/documentation/build/paths#writing-javascript-expressions-for-custom-path-conditions

josephjclark:
Closes #451  I'd say, especially with a companion PR coming up in Lightning (EDIT: https://github.com/OpenFn/lightning/pull/2238)

I haven't addressed troubleshooting at all because frankly the experience is so bad. We need a bit of development effort here first. Hopefully if we can produce good devx we won't even need docs.

### DIFF
--- a/docs/build/paths.md
+++ b/docs/build/paths.md
@@ -9,34 +9,58 @@ and configuration tips.
 
 ## Path Conditions
 
-There are 4 types of Path Conditions that define whether the Workflow will
+There are four types of Path Conditions that define whether the Workflow will
 proceed to the next Step when executed:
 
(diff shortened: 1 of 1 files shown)


josephjclark:
Also, we're doing this in Lightning now:

aleksa-krolls (approved):
Thanks, good to have some examples @josephjclark !

PR #513, opened 2024-07-02
Fix links in triggers page

josephjclark:
Here's what [https://docs.openfn.org/documentation/build/triggers](https://docs.openfn.org/documentation/build/triggers) looks like right now:

image

The "Managing State" and "Keep a cursor" sections don't link anywhere and aren't on the page.

This was pointed out by a user over in https://github.com/OpenFn/adaptors/pull/654#issuecomment-2200916417

Looking at the legacy docs it looks things have been moved around and these got left hanging.

So I've made a few changes:
* Link the cursor stuff to the job writing guide bit on cursors
* Link to some state management stuff for triggers (although to be fair I don't think it's a very useful link)
* Removed some bullet points which I didn't think were very helpful
* Used Proper Nouns instead of "quotes"  for jargon like Webhook Event, which looks a little bit less like we're making up words
* Adjusted the  page structure to be a bit more useful

### DIFF
--- a/docs/build/triggers.md
+++ b/docs/build/triggers.md
@@ -2,20 +2,19 @@
 title: Triggers
 ---
 
-Triggers allow you to start the executiong of workflows automatically. They come
-in two types: "cron" triggers and "webhook event" triggers.
+Triggers allow you to start the execution of workflows automatically. They come
+in two types: Cron triggers and Webhook Event triggers.
 
-## Trigger types
-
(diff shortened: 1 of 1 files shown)


josephjclark:
@taylordowns2000 @aleksa-krolls  maybe either of you could check and merge this please :)

taylordowns2000 (approved):

PR #514, opened 2024-07-02
Odk overview

josephjclark:
Picking up Yaw's contribution to the ODK overview.

I don't really know what we're doing with overviews but we do need to update the copy we have right now.

It would be nice to include an example of fetching all form data and building all submissions for a project into an array. Or grouping forms.

### DIFF
--- a/adaptors/odk.md
+++ b/adaptors/odk.md
@@ -1,8 +1,16 @@
 ---
-title: ODK
+title: Open Data Kit (ODK)
 ---
 
-In order to fetch data from ODK Central, you must run a scheduled job with a
+# App Overview
+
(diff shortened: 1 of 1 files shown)


yanokwa:
Could you add little bit of context at the top? Any blockers to merging I can help with?

># App Overview
>
>[ODK](https://getodk.org) is a data collection platform used by researchers, field teams, and other professionals to collect they data they need wherever it is.
>
># Integration Options
>
>In order to fetch data from an ODK Central server, you must run a scheduled job with a cron trigger using the `odk` adaptor.
>
>Here's an example of a job that gets submission data....

josephjclark:
@yanokwa Done - we'll get this released today. Thanks!

PR #518, opened 2024-07-17
Add documentation for promises

josephjclark:
We've just released support in the runtime to treat every operation as a Promise.

See https://github.com/OpenFn/kit/pull/722 for all the gory details.

The tl;dr on this is:

1) You can now do `fn().catch()`
2) `.then()` should remove the need for 95% of callbacks in adaptor code
3) You should start doing `each($.data, fn().then())`

This is a really big deal and needs good documentation.

The tension here  is that I want to communicate the change to users who know openfn, but to new users this promises and callbacks thing is an edge case really.

In the future, most of the existing Callbacks section of the docs will be removed, and we'll just have this Promises stuff and maybe a section explaining `fn()` (because a lot of callbacks in existing job code should now migrate to using an fn block)

### DIFF
--- a/docs/jobs/job-writing-guide.md
+++ b/docs/jobs/job-writing-guide.md
@@ -90,7 +90,129 @@ Your job code should only contain Operations at the top level/scope - you should
 NOT include any other JavaScript statements. We'll talk about this more in a
 minute.
 
+## Promise-like Operations
+
+:::tip
+
+Promise support is new to the openfn since July 2024, introduced in
+`compiler@version` and `runtime@version`
+
+:::
+
+Operations behave like Javascript Promises in that they have `.then()` and
+`.catch()` functions. This is useful for creating your own callbacks and error
+handling
+
+### Callback with then()
+
+`then()` is available on every operation. It accepts the state as its argument
+and MUST return a state object to be passed to the _next_ operation.
+
+```js
+fn().then(state => {
+  console.log(state);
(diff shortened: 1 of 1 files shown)


aleksa-krolls:
@ritazagoni @mtuchi can you pls review and provide feedback here?

josephjclark:
I've just updated the docs to be a bit sharper. I've also moved it all down the page.

The docs  should be self explanatory. Eventually this section should replace the `Callbacks` bit entirely.

docs/jobs/job-writing-guide.md
```diff
@@ -568,6 +577,146 @@ you how the compiler is treating your State operators.
...
+:::tip
+
+You can read more about the `each()` operation in
+[Iteration with Each](#iteration-with-each).
+
+:::
+
+The `each` function will take an array and, for each item, invoke a callback
+with a scoped state. This means it takes your state object and sets the item
+under iteration to `state.data`. In order words, `state.data` inside the
```
ritazagoni, line 651: Would this be "in other words"?

ritazagoni (approved):
Left one comment for possible typo, otherwise looks great and very clear!

docs/jobs/job-writing-guide.md
josephjclark: Ooh good catch, thanks!

PR #522, opened 2024-07-26
Configurable Concurrency limit by workflow

christad92:


taylordowns2000 (approved):
note that i moved this into the workflows page and added a few examples. all good otherwise!

PR #515, opened 2024-07-08
ms graph adaptor docs

ritazagoni:

PR #526, opened 2024-07-30
Overview page for the MSGraph Adaptor

christad92:


ritazagoni (approved):

aleksa-krolls (changes requested):
hey @christad92 @ritazagoni what is going on in this PR? 
1. Firstly why are we updating the MS Graph overview page in this PR _and_ in [this other PR](https://github.com/OpenFn/docs/pull/515)
2. This PR also contains a load of other file changes I do _not_ think we want to merge (e.g., overwriting ODK page, deleting roadmap page, etc.). I suspect this should be a mistake, but we should break up these other changes into separate PRs

PR #531, opened 2024-08-07
Yarn 3.8.3

josephjclark:
Hey @taylordowns2000 I wanna get this yarn stuff wrapped. I really don't understand what's going on but I think this PR will get us some parity.

This PR locks yarn to 3.8.3. Locally I am using 3.8.3.

There is a diff in the lockfile but it's tiny, a couple of changed chache keys and checksums. I don't understand it but it's small enough that I don't care.

Note that if we use the latest yarn, 4.x, the lockfile changes version and we get all sorts of changes.

I don't want any of that: I just want to lock the yarn version down with minimal changes.  I think this does it.

Could you please check out this branch and run `yarn` from the root?

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -14,6 +14,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
+      - run: corepack enable
       - uses: actions/setup-node@v4
         with:
           node-version: 20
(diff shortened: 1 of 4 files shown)


josephjclark:
Running a test on my second machine

In an old version of the docs repo, it is running yarn 1.22.22 (just like my main machine actually)

When I run `yarn` on this branch, I get an error because corepack isn't enabled. Cool.

Now `yarn` runs properly and takes ages.

When it finishes I have no diff in my local repo.

I'm happy with that

PR #533, opened 2024-08-11
Showcase support for file paths in `project-spec.yaml`

midigofrank:
The spec now supports 2 ways of defining job bodies: inline, as or as an object referencing an external file.

Depends on: https://github.com/OpenFn/kit/pull/742


taylordowns2000 (approved):
Fantastic. Will hold merge until the other is released. Thanks!

PR #543, opened 2024-08-23
Clean up and organize list of commonly used CLI features in cli-usage.md

DrCord:
Clean up and organize list of commonly used CLI features. Clarify link, remove "here" as best practice for linking online.

### DIFF
--- a/docs/build-for-developers/cli-usage.md
+++ b/docs/build-for-developers/cli-usage.md
@@ -4,7 +4,7 @@ sidebar_label: Basic usage
 slug: /cli-usage
 ---
 
-This page shows examples for some of the most common usages of the CLI, such as:
+This page shows examples for some of the most common usages of the CLI, including:
 
 - get help
 - run a job
@@ -30,8 +30,7 @@ openfn deploy --help
 
 ### Run a job
 
-To run a single job, you must explicitly specify which adaptor to use.
-[Publicly available adaptors](/adaptors). See examples below.
+To run a single job, you must explicitly specify which adaptor to use - see the [publicly available adaptors](/adaptors).
 
 Adaptors are automatically installed if the specified version is not detected.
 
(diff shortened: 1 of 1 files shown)


docs/build-for-developers/cli-usage.md
```diff
@@ -4,10 +4,15 @@ sidebar_label: Basic usage
 slug: /cli-usage
 ---
 
-This page shows common usage examples for the CLI.
+This page shows examples for some of the most common usages of the CLI, such as:
```
josephjclark, line 7: Yes, this is an improvement for sure. Can we maybe say `including:` instead of `such as:`? 

To be fair this stuff is also in the side nav menu, so I wouldn't be against removing this whole intro paragraph :shrug:  It's not adding much is it?

docs/build-for-developers/cli-usage.md
```diff
@@ -25,9 +30,8 @@ openfn deploy --help
 
 ### Run a job
 
-To run a single job, you must explicitly specify which adaptor to use. You can
-find the list of publicly available [adaptors here](/adaptors). See examples
-below.
+To run a single job, you must explicitly specify which adaptor to use.
```
josephjclark, line 33: Can I suggest we remove `see examples below` and have a less terse link?

Maybe this:
```
To run a single job, you must explicitly specify which adaptor to use - see the [publicly available adaptors](/adaptors).
```

docs/build-for-developers/cli-usage.md
```diff
@@ -173,8 +177,8 @@ openfn repo clean
 
 ### Run a workflow
 
-<details>
-  <summary>A workflow has a structure like this</summary>
+<details open>
```
josephjclark, line 180: hmm. I take your point but I'd probably rather hide the workflow json, it's a bit big. I can't even remember why we're including it.

Can we instead rename the summary text to `Click to see a workflow's JSON structure` or something?

josephjclark (changes requested):
Couple of discussion points on this one I think @DrCord. I agree with the problems you've identified - just not sold on the solutions in this case!

josephjclark (approved):
Thank you!

PR #545, opened 2024-08-26
Updated docs on Github/Portability

christad92:
Updated docs on GitHub sync
Connected/Referenced both docs where context is required


aleksa-krolls (approved):
Thanks @christad92 ! This is helpful :)

PR #550, opened 2024-08-26
Fix awkward phrasing paths.md

DrCord:
add a missing word to fix awkward phrasing


josephjclark (approved):
Thanks, that's better!

PR #555, opened 2024-08-26
Fix typos in activity-history.md

DrCord:


taylordowns2000 (approved):
worka -> works but I'll make that change now. Thank you!

PR #556, opened 2024-08-26
Clarify mappings example in troubleshooting.md

DrCord:


josephjclark (changes requested):
I'm not so sure on this one? I see the problem but there are no two "For examples" in this paragraph.

Maybe if we tweak (mostly remove) the sentence before, it'll flow a bit more clearly

I will come back and look at this next week

josephjclark (approved):

josephjclark:
@DrCord Thanks!

PR #561, opened 2024-08-27
Add missing link to internal docs on manage-credentials.md

DrCord:


christad92 (approved):

taylordowns2000 (changes requested):
thank you! this is great but ill reject to use `/documentation/blah` rather than the external site link.

PR #599, opened 2024-11-07
Collections Docs

josephjclark:
## Short Description

Docs for Collections:

* Under Platform / Build & Manage Workflows we have a high-level Collections overview. This has a couple of code examples but is designed as a user guide, rather than a developer / job writing guide
* Under Adaptors / Collections is a detailed API overview and explanation of how it all works (the important bits anyway)

There is a polite notice indication that collections are in Beta (ie , not totally production ready right now, thanks, but have a play)

## AI Usage

know!):

- [ ] Strategy / design
- [ ] Other
- [x] I have not used AI


adaptors/collections.md
```diff
@@ -0,0 +1,200 @@
+---
+title: Collections Adaptor
+---
+
+## Collections Overview
+
+The Collections API is a key/value storage solution. It is designed for high
```
aleksa-krolls, line 7: We need to make clear that this is a data store _in_ OpenFn... not something external (like all other adaptors).

adaptors/collections.md
```diff
@@ -0,0 +1,200 @@
...
+performance over a large volume of data.
+
+Use-cases include:
+
+- Storing mapping objects for use in workflows
+- Buffering and aggregating high volumes of incoming data
+- Caching and sharing state between workflows
+
+A Collection is bound to a project. Collections can only be accessed with a
```
aleksa-krolls, line 16: "a project in OpenFn"?

adaptors/collections.md
```diff
@@ -0,0 +1,200 @@
...
+token associated with that project. When running on the app, a workflow is
```
aleksa-krolls, line 17: You mean a user's API Key? If yes, we should link to the docs instructions on how to generate one
https://docs.openfn.org/documentation/api-tokens

adaptors/collections.md
```diff
@@ -0,0 +1,200 @@
+---
+title: Collections Adaptor
```
aleksa-krolls, line 2: Wondering if we can somehow tag `collections` and `common` as special OpenFn adaptors?

adaptors/collections.md
```diff
@@ -0,0 +1,200 @@
...
+under a given key:
+
+You can also pass an array of items for a batch-set.
+
+### Getting data from a collection
+
+To retrieving multiple items from a Collection, we recommend using the `each()`
+function.
+
+`each()` will stream each value individually, greatly reducing the memory
```
aleksa-krolls, line 45: Do we really have to use `each()`? What about the scenario when I want to query a list of devices that match my search criteria? In some of the use cases we've been imaging so far, we've been getting a long list of records and then sending that entire list to a downstream system, or comparing that list with data from another system. Mtuchi could show you what's going on the openlmis prototype, if interested in learning more.

adaptors/collections.md
```diff
@@ -0,0 +1,200 @@
...
+  state.resources ??= {};
+  state.resources[state.data.id] = mappedString;
+  return state;
+});
+```
+
+You can also fetch multiple items with `get()`, which supports the same query
+options as `each()`.
+
+Bear in mind that all the items will be loaded into memory at once. For large
```
aleksa-krolls, line 102: Hmm... will ping you next week to help me understand tradeoffs of using `get()` and `each()` for different use cases so I can better understand when we might start bumping into limits, and to know which is best for different scenarios... we can then beef up these docs with some examples.

adaptors/collections.md
aleksa-krolls: I see `get()` now and left a comment below

aleksa-krolls (changes requested):
Major feedback is to make it clear that `collections` is something _in_ openfn and not an external adaptor. 

Then generally, services team can help add more examples and use cases that will help users better imagine how to use this... but we can do that later and debate where those docs go.

taylordowns2000 (commented):
Looks good, but I wanted to try out collections on staging just now and was hoping to see how to add items to a collection. Is that documented somewhere else?

josephjclark:
@taylordowns2000 My bad - I forgot to add set values. That's coming up in a minute

adaptors/collections.md
josephjclark: Don't ping me - let me fix it here!

adaptors/collections.md
josephjclark: Kinda - thanks for flagging, I'll clean this up

adaptors/collections.md
josephjclark: Maybe. Collections is definitely special. Common is only sort of special :thinking: 

So I guess to answer this I'd have to ask what we mean by "special" and what exactly we want to flag

josephjclark:
@aleksa-krolls So  to be clear (and the context is probably a bit lost), this is the Overview adaptor docs. If you go to the doc site and the adaptor list, this is the Overview. It's designed as a high level introduction to the collections  feature generally and the collections API specifically.

Collections support is implemented as an adaptor. That's how we inject an API in job code, and how we provide documentation. We should not hide or obfuscate this because it's kind of relevant - the collections adaptor appears in run logs, and if you're using the CLI, you need to pass the collections adaptor explicitly for support.

I agree that we need some more general, less technical docs somewhere in the  main docs page. I'm not sure where. I suppose there's a general `Use Collections` style page which explains the feature at a high level (this can show API examples and will link to the adaptor docs, but otherwise doesn't treat collections as an adaptor), plus a `Collections Admin` page which explains how to create, delete and a generally manage collections through the admin API.

But I don't know WHERE that goes. Maybe under Platform as a section called Collections? Or nested under Platform / Build & Manage Workflows (that seems very deeply nested, but I think it makes sense as a peer to eg Credentials, Limits, Snaphots, Triggers)

aleksa-krolls:
@josephjclark I agree it makes sense there. Can we do that to start, and can you please raise in engineering standup to get this done sometime in the coming days? This page should then link to the adaptor docs. 
> But I don't know WHERE that goes. Maybe under Platform as a section called Collections? Or nested under Platform / Build & Manage Workflows (that seems very deeply nested, but I think it makes sense as a peer to eg Credentials, Limits, Snaphots, Triggers)

Screenshot 2024-11-07 at 9 05 57 PM

josephjclark:
@aleksa-krolls I'm taking responsibility for docs on this, so I'll get it done tomorrow.

josephjclark:
So:

* The Adaptors page is now a technical overview and API reference for the Collections API.
* The Docs page is a high level overview of the feature. It has a couple of code examples and refers to the API, but otherwise is a conceptual overview and usage guide more than a technical one.

@stuartc Would you mind looking over both pages and calling out any egregious errors?

These docs DO NOT include:

* performance tips for big data sets (eg, use each not get). I would like to do this later but I just don't have time and I'm cutting cloth
* Details on usage limits and costs. I have no idea what the rules are here
* A good sample workflow which uses collections. Not that we have a good place to share this at the moment...

Please also note that I've raised an issue for better collections support in the CLI (eg, `openfn collections set` would be awesome), se https://github.com/OpenFn/kit/issues/819

docs/build/collections.md
```diff
@@ -0,0 +1,150 @@
...
+sidebar_label: Collections
+---
+
+Collections provides a high-volume, high-performance storage solution built into
+OpenFn.
+
+Collections is suitable for buffering, caching and aggregating data from
+Webhooks, storing large mapping files, and sharing state between workflows.
+
+Collections can be used to store a very large number of items (in the order of
```
josephjclark, line 12: @stuartc your nervousness please about the phrase `in the order of millions`

I just want give some idea of what "high volume" means. I know it's woolly.

stuartc: I think we can leave it as is for now. We will need to determine two things:

1. How much work it is for the db to store really large values, right now they are strings and not indexed, but can be really large; save for our middleware that will reject large (>10MB) HTTP requests.
2. If users are ok without being able to query by value.

If either of these are an issue, then the number of records we would want to limit would be clearer. But for now, millions should be ok.

stuartc (approved):
Wonderful! 🫶

aleksa-krolls (approved):

PR #606, opened 2024-11-27
November roadmap update

christad92:
## Short Description

Monthly roadmap changes with update to reflect moving away from Canny to Community

## Details

Add any more details that may be relevant to the reviewer. How did you approach
these docs? Are there any parts you struggled with, or that need particularly
careful review?

## AI Usage

know!):

- [ ] Strategy / design
- [x] Other - I use Cursor.com as my editor
- [ ] I have not used AI


taylordowns2000 (changes requested):
@christad92 , this looks great! Especially if we're only listing out _big stuff_ here, I think it might be useful to keep the current _and_ previous quarter.

So if this update comes in November, we'd want Q3'24, Q4'24 (current), and Q1'25.

Can you leave Q3 in there?

Then, when you do an end of January update you can have Q4'24, Q1'25, Q2'25.

christad92:
Thanks @taylordowns2000 i have added the Q3 stuff in a new commits. Please take a look and let me know if it looks good

taylordowns2000 (changes requested):
Perfect. Thank you. Just realized one more thing.... the features should not be numbered if the numbers will change every time you make a commit. Remove those numbers at the start of the feature name and the diffs will look way better!

PR #651, opened 2025-03-20
satusehat: add job examples

hunterachieng:
## Short Description

Add job examples to the `satusehat` adaptor

Closes [#971](https://app.zenhub.com/workspaces/implementation-engineering-600030af88f2bb00155d46a5/issues/gh/openfn/adaptors/971)

## Details

Add any more details that may be relevant to the reviewer. How did you approach
these docs? Are there any parts you struggled with, or that need particularly
careful review?

## AI Usage

know!):

- [ ] Strategy / design
- [ ] Other
- [ ] I have not used AI


josephjclark (changes requested):
Can we try and do something a bit more narrative in these examples?

These single-statement examples don't really show much more usage than the docs examples.

And actually for satusehat, what we really want to show in examples is how to take data from local health system, convert it into fhir, and then upload it. That's a really really valuable process to show to users out here with real problems.

Ok, so we  can't do a whole workflow, but I'm sure we can simulate one. Or just cover the hard bits.

Maybe @AishaHassen can help you tell this convert-to-fhir story in a single example? I want to see some mapping  code examples, some state manipulation, some fn blocks and maybe some each.

We don't necessarily need to see a whole resource getting built (because that's a  lot of data) but it would be nice to see the whole pipeline for part of a resource. Like just mapping an identifier and address or something.

adaptors/library/jobs/satusehat-create-medication.js
```diff
@@ -0,0 +1,134 @@
+// Create a medication from visit data and lookup table
+
+function generateUUID() {
```
josephjclark, line 3: Not keen on spreading this pattern! It's not a nice function at all

Common should have a util.uuid. Can we use that instead? If it's not exported by satusehat, let's do a quick patch (just make sure common.util is exported by satusehat)

hunterachieng: @josephjclark  We are not exporting the `util.uuid` from `common` in satusehat

josephjclark: Then let's do a patch and do that :) Should only take a minute an it'll make this example significantly cleaner

I've only had a light look over this btw - I'll get back to you on the rest of it later

josephjclark:
@hunterachieng so is this literally just a copy and paste of an existing workflow?

My worry with this is that it's still not really telling the story in a very clear way. Like the first line of create medication is:
```
fnIf(
  !state.errors &&
    state.visit.properties['prescription_1'] &&
    state.visit.properties['prescription_1'].length > 0,
```
But like, what have all those conditions got to do with anything?

That's part of the _workflow_ story. But here we only have a single step.

For this example, we can probably assume that `state.visit` is what we need.

But we need  a little bit of commentary in comments, like: `state.visit is a Commcare Visit record`. And then the step name is something like "Create a FHIR Encounter from a Commcare Visit"

Our job isn't to capture a whole workflow, but to show how a non-fhir object might be converted into fhir, and then uploaded to Satusehat. An even then we don't need to do a whole commcare record - just enough to show the sort of patterns needed.

For example: the mapMedications function is really interesting.  It shows how we can declare helper functions in  job code, it shows how arbitrary commcare strings can be mapped into proper FHIR resources, it shows how to do looping in a functional programming style.

`createMedicationResource` is also super interesting, because it shows how you can create all the FHIR boiler-plate and map resources into it. It shows a technique for building a fhir object. This is also super useful!

So we need to priorise what's important and interesting about the example, document it at a high level so that users can understand it, and clear out anything that's noisy or confusing

adaptors/library/jobs/satusehat-create-encounter.js
```diff
@@ -0,0 +1,105 @@
+// Create an encounter in Satusehat
+
+fn(state => {
+  state.organizationId = state.visit.organizationId;
+  state.encounterId = util.uuid();
+
+  state.encounterMapping = {
+
```
AishaHassen, line 8: hey @hunterachieng can you format ur code with prettier? i see a lot of empty lines all over.

adaptors/library/jobs/satusehat-create-encounter.js
```diff
@@ -0,0 +1,105 @@
+// Create an encounter in Satusehat
+
+fn(state => {
+  state.organizationId = state.visit.organizationId;
```
AishaHassen, line 4: can we remove this line and just use, `state.visit.organizationId` when constructing your encounter?

adaptors/library/jobs/satusehat-create-observations.js
```diff
@@ -0,0 +1,144 @@
+// Create an observation in Satusehat
+
+fn(state => {
+  function createObservationData(item, state, result) {
+    const patientReference = `Patient/${
+      state.patient.ihs_number ?? state.patient.properties.ihs_number
```
AishaHassen, line 6: Can you simplify this as `state.patient.ihs_number` remove `?? state.patient.properties.ihs_number`

adaptors/library/jobs/satusehat-create-observations.js
```diff
@@ -0,0 +1,144 @@
...
+
+          if (result) {
+            const data = createObservationData(item, state, result);
+            dataObjects.push(data);
+          }
+        }
+
+        return dataObjects;
+      })
+      .filter(data => data !== null);
```
AishaHassen, line 131: not sure what this line is doing, can u remove it?

adaptors/library/jobs/satusehat-update-encounter.js
```diff
@@ -0,0 +1,77 @@
+// Update Encounter resource with diagnosis data
```
AishaHassen, line 1: I say we remove this example. it is too complicated to understand without ASRI context and it will go over people's heads

AishaHassen (commented):
Hey @hunterachieng , we're missing the example for conditions like we talked about yesterday. General comments for all the js functions:
1. add comments wherever relevant
2. use prettier when you are done

adaptors/library/jobs/satusehat-create-condition.js
```diff
@@ -0,0 +1,97 @@
+// Create a diagnosis in satusehat
+
+fn(state => {
+  state.conditionMappedData = state.visit.diagnosis.flatMap(diagnosis =>
```
josephjclark, line 4: I would probably call this `state.conditions`. I can guess what a conditions array is - but I'm not sure what a `conditionMappedData` thing is

adaptors/library/jobs/satusehat-create-condition.js
```diff
@@ -0,0 +1,97 @@
+// Create a diagnosis in satusehat
+
+fn(state => {
```
josephjclark, line 3: Maybe a comment like "Map diagnosis objects [from commcare?] into Condition resources", just to set some context

adaptors/library/jobs/satusehat-create-condition.js
```diff
@@ -0,0 +1,97 @@
...
+  state.conditionMappedData = state.visit.diagnosis.flatMap(diagnosis =>
+    state.lookup_table.dentist_diagnosis
```
josephjclark, line 5: What is `state.lookup_table` and where did it come from? It feels like an important thing to explain. Or at least mention?

adaptors/library/jobs/satusehat-create-condition.js
```diff
@@ -0,0 +1,97 @@
...
+      .filter(item => diagnosis === item.fields.name)
```
josephjclark, line 6: Can we use a better name than `item` here? The item is a diagnosis in the lookup table right?

adaptors/library/jobs/satusehat-create-condition.js
```diff
@@ -0,0 +1,97 @@
...
+        };
+      })
+  );
+  return state;
+});
+
+post('Condition', $.conditionMappedData);
+
+fn(state => {
+  // Storing the condition resources in state
```
josephjclark, line 94: Maybe `// Store the newly created resources, created by the server, into state for the next step`

adaptors/library/jobs/satusehat-create-condition.js
```diff
@@ -0,0 +1,97 @@
...
+          },
+          onsetDateTime: state.visit.visit_date,
+          recordedDate: state.visit.visit_date,
+        };
+      })
+  );
+  return state;
+});
+
+post('Condition', $.conditionMappedData);
```
josephjclark, line 91: This line is pretty obvious but I'd still say "Post our condition resources to Satusehat`

adaptors/library/jobs/satusehat-create-encounter.js
```diff
@@ -0,0 +1,101 @@
+// Create an encounter in Satusehat
```
josephjclark, line 1: Maybe this should be like `Create an Encounter  in Satusehat, based on an incoming Commcare visit`

We don't have to explain the Commcare visit structure or anything, but it'll really help people understand what's happening if they have some frame of reference for what the input structure represents.

I suppose that context could be set by the example name too

adaptors/library/jobs/satusehat-create-medication.js
```diff
@@ -0,0 +1,115 @@
+// Create a medication from visit data and lookup table
+
+fn(state => {
+  // state.visit is a Commcare Visit record
```
josephjclark, line 4: Yeah nice - this really helps I think!

adaptors/library/jobs/satusehat-create-medication.js
```diff
@@ -0,0 +1,115 @@
...
+  const properties = state.visit.properties;
+  const medicineKeys = extractMedicineKeys(properties);
+
+  const obatList = state.lookup_table.obat;
+
+  state.medicationSatusehatId = [];
+
+  // Map medication data for each relevant key
+  state.medicationMappedData = mapMedications(
```
josephjclark, line 13: Still not sure about the `MappedData` suffix. I can see it's a repeated pattern so if it's clear for other users, then ignore me :)

adaptors/library/jobs/satusehat-create-medication.js
```diff
@@ -0,0 +1,115 @@
...
+      resourceType: 'Medication',
+      id: state => state.medicationId,
+      meta: {
+        profile: [
+          'https://fhir.kemkes.go.id/r4/StructureDefinition/Medication',
+        ],
+      },
+      identifier: [
+        {
+          system: `http://sys-ids.kemkes.go.id/medication/${state.asri_satusehat_id}`,
```
josephjclark, line 83: Weird that this is using state. Looks like a bug? Maybe we need to pass this value into the function?

adaptors/library/jobs/satusehat-create-medication.js
```diff
@@ -0,0 +1,115 @@
...
+function createMedicationResource(item, medicationId) {
+  return {
+    fullUrl: `urn:uuid:${medicationId}`,
+    request: {
+      method: 'POST',
+      url: 'Medication',
+    },
+    resource: {
+      resourceType: 'Medication',
+      id: state => state.medicationId,
```
josephjclark, line 75: is this `state.medicationId` different to `medicationId` ? I don't think so?

adaptors/library/jobs/satusehat-create-observations.js
```diff
@@ -0,0 +1,140 @@
+// Create an observation in Satusehat
+
+fn(state => {
+  function createObservationData(item, state, result) {
```
josephjclark, line 4: I don't think the two functions here should be declared inside the `fn` block. Looks like a strange pattern to me

adaptors/library/jobs/satusehat-create-observations.js
```diff
@@ -0,0 +1,140 @@
...
+      },
+      effectiveDateTime: state.visit.visit_date,
+      ...result,
+    };
+  }
+
+  /**
+   * Filters and maps observations based on the visit.
+   */
+  function mapObservations(state) {
```
josephjclark, line 40: As `mapObserverations` is only called once, should this really be in its own function?

adaptors/library/jobs/satusehat-create-observations.js
```diff
@@ -0,0 +1,140 @@
...
+
+  state.observationMappedData = mapObservations(state);
+
+  return state;
+});
+
+post('Observation', $.observationMappedData);
+
+fn(state => {
+  state.observatioinData = state.data;
```
josephjclark, line 138: There's a typo here on `observatioinData`

adaptors/library/jobs/satusehat-create-observations.js
```diff
@@ -0,0 +1,140 @@
...
+  function mapObservations(state) {
+    return state.visit.observations.map(item => {
+      const { value_type: valueType, value } = item;
+      const dataObjects = [];
+
+      if (valueType === 'valueCodeableConcept') {
+        value.forEach(value => {
+          const result = {
+            [valueType]: {
+              coding: [JSON.parse(value.snomed_code)],
```
josephjclark, line 49: Curious as to why we have all this JSON parsing? Is that really needed? Can we cheat in this example and pretend the input is well formatted?

adaptors/library/jobs/satusehat-create-observations.js
```diff
@@ -0,0 +1,140 @@
+// Create an observation in Satusehat
+
+fn(state => {
```
josephjclark, line 3: I'd love to see this implemented as an `each` rather than an `fn`. But  I think actually you'd want this to be a `map()` right? Take one array and map the contents into a new one? So let's not change anything but I've made a mental note of this use-case

josephjclark (commented):
Left a few comments but maybe I'm being too picky? It's come on a long way :)

adaptors/library/jobs/satusehat-create-observations.js
hunterachieng: I think we need to show them how we create the different mappings since it is an important aspect when creating an obervation. cc @AishaHassen

AishaHassen (approved):
@hunterachieng all good to merge on my end once you resolve joe's comments.

adaptors/library/jobs/satusehat-create-observations.js
hunterachieng: @josephjclark  I have removed the parsing

josephjclark (approved):

PR #669, opened 2025-04-22
add update trigger for docsite embeddings

hanna-paasivirta:
## Short Description

Add workflow action to trigger an update to docsite embeddings in Apollo when the docsite is updated.

## Details

Posts a request to Apollo to trigger the embed_docs service. It requires Pinecone and OpenAI keys (to be added in secrets).

## AI Usage

know!):

- [x] Code generation (copilot but not intellisense)
- [x] Learning or fact checking
- [x] Strategy / design
- [x] Translation / spellchecking / doc gen
- [ ] Other
- [ ] I have not used AI

### DIFF
--- a/.github/workflows/trigger-embed.yml
+++ b/.github/workflows/trigger-embed.yml
@@ -1,0 +1,22 @@
+name: Trigger Embed Service
+
+on:
+  push:
+    branches:
+      - main
+
+jobs:
+  trigger-embed:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Trigger Embed Service
+        env:
+          PINECONE_API_KEY: ${{ secrets.PINECONE_API_KEY }}
+          OPENAI_API_KEY: ${{ secrets.OPENAI_API_KEY }}
+        run: |
+          curl -X POST https://apollo.staging.openfn.org/services/embed_docsite \
+            -H "Content-Type: application/json" \
+            -d '{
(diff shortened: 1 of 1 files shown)


.github/workflows/trigger-embed.yml
```diff
@@ -0,0 +1,22 @@
...
+jobs:
+  trigger-embed:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Trigger Embed Service
+        env:
+          PINECONE_API_KEY: ${{ secrets.PINECONE_API_KEY }}
+          OPENAI_API_KEY: ${{ secrets.OPENAI_API_KEY }}
+        run: |
+          curl -X POST https://apollo.staging.openfn.org/services/embed_docsite \
```
josephjclark, line 17: We don't have production and staging pinecone databases, right? We only have one?

So calling out to staging will update the shared database, and production will also get the latest versions on the next search?

hanna-paasivirta: As long as the key that is added on GitHub for this action is the same as the key used in Apollo in the AI Assistant then it will be the same shared database. So we need to make sure we add the same one on GitHub – do you have permission to add secrets?

josephjclark: Perfect. Maybe we can add a comment here because this might confuse someone later on

Keys are all sorted!

josephjclark (approved):

josephjclark:
wonderful - holding off the merge until staging is ready

PR #765, opened 2026-04-28
add docs for configuring webhook responses for Sync triggers

midigofrank:
## Short Description

add docs for configuring webhook responses for Sync triggers

## AI Usage

know!):

- [x] I have used Claude Code
- [ ] I have not used AI

[Responsible AI Policy](https://www.openfn.org/ai)

### DIFF
--- a/docs/build/triggers.md
+++ b/docs/build/triggers.md
@@ -47,6 +47,9 @@ gets a fast acknowledgement and the workflow runs in the background.
 **Response:**
 
 - Status code: `200`
+- Headers:
+  - `x-meta-work-order-id` — the ID of the created work order
+  - `x-meta-run-id` — the ID of the created run
 - Body:
   ```json
   {
@@ -71,6 +74,9 @@ system waits — sometimes seconds or minutes — for the result.
 **Default response:**
 
 - Status code: `201` (configurable — see below)
+- Headers:
+  - `x-meta-work-order-id` — the ID of the work order
+  - `x-meta-run-id` — the ID of the run
 - Body: the final run state as a JSON object
 
 :::note Security policy for failed runs
--- a/docs/build/triggers.md
+++ b/docs/build/triggers.md
@@ -27,34 +27,24 @@ Learn how a workflow's initial `state` gets built from a webhook trigger
 
 ## **Webhook Trigger Responses**
 
-When a workflow is triggered via a webhook, OpenFn can run that workflow in one
-of two ways, depending on how the trigger is configured and what the calling
-system needs.
+When a workflow is triggered via a webhook, OpenFn can respond to the calling
+system in one of two ways, depending on how the trigger is configured.
 
-### **Asynchronous mode**
+### **Async (Before Start)**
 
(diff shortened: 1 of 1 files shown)


docs/build/triggers.md
```diff
@@ -27,36 +27,29 @@ Learn how a workflow's initial `state` gets built from a webhook trigger
...
 ## **Webhook Trigger Responses**
 
-When a workflow is triggered via a webhook, OpenFn can run that workflow in one
-of two ways, depending on how the trigger is configured and what the calling
-system needs.
+When a workflow is triggered via a webhook, OpenFn can respond to the calling
+system in one of two ways, depending on how the trigger is configured.
 
-### **Asynchronous mode**
+### **Async (Before Start)**
```
josephjclark: I see what you're getting at, but can I recommend the title is  `Async Mode Responds Before Start`

Just because the raw option text doesn't really have enough context to make this make sense

docs/build/triggers.md
```diff
@@ -65,53 +58,89 @@ run. It does **not** include the workflow’s output.
...
 
-After the workflow completes, sometimes seconds (or minutes) later.
+- Status code: `201` (configurable — see below)
+- Headers:
+  - `x-meta-work-order-id` — the ID of the work order
+  - `x-meta-run-id` — the ID of the run
+- Body: the final run state as a JSON object
 
-**What this response includes:**
+:::note Security policy for failed runs
```
josephjclark, line 82: Great 👍

docs/build/triggers.md
```diff
@@ -65,53 +58,89 @@ run. It does **not** include the workflow’s output.
...
-      "state": "success",
-      "error_type": null,
-      "inserted_at": "2025-10-23T10:15:00Z",
-      "started_at": "2025-10-23T10:15:05Z",
-      "claimed_at": "2025-10-23T10:15:06Z",
-      "finished_at": "2025-10-23T10:15:20Z"
-    }
-  }
-  ```
+In sync mode you can set custom HTTP status codes via the trigger inspector
```
josephjclark: Is "trigger inspector" clear? Do we use that language elsehwere?

Is it better to just say "set custom HTTP status codes on the trigger" ? That sort of covers the UI and CLI

docs/build/triggers.md
```diff
@@ -65,53 +58,89 @@ run. It does **not** include the workflow’s output.
...
+fn(state => ({
+  ...state,
+  webhookResponse: {
+    status: 200,
+    body: { ack: true, id: state.data.id },
+  },
+}));
+```
+
+`webhookResponse` is captured from each step as it completes, with **last write
```
josephjclark: This is a bit too technical. Although this might not be technically accurate, I think I'd say:

```
At the end of the run, the value of `state.webhookResponse` will be used to send the HTTP response back to the caller. Changing the value during the run does not affect the response: it's only the final state that counts.
```

josephjclark (commented):
Lovely, just a couple of comments

josephjclark (approved):

PR #790, opened 2026-05-18
Sandboxes: corrections, new sections, and screenshots

elias-ba:
## Summary

Rewrites `docs/build/sandboxes.md` against the current `release/2.16.4` behaviour in Lightning.

## Corrections

- **Credentials are shared with the parent**, not duplicated. `Sandboxes.clone_credentials_from_parent/2` only copies the `project_credentials` link; the credentials themselves are shared.
- **Subscription / billing is shared**, not separate. `Subscription` `belongs_to :project` is constrained to root projects (`sandboxes share their parent's subscription`), so usage in a sandbox counts toward the parent's allowances.
- **Triggers start disabled at creation.** `clone_triggers_with_disabled_state/2` sets `enabled: false` unconditionally.
- **"Destroyed after merging" is wrong.** Source is scheduled for deletion with the configured grace period; it's restorable from the sandboxes list during the window.
- **Conflicts section** described a non-existent "force the merge" toggle. Actual UX is per-workflow `Changed` / `Diverged` / `New` / `Deleted in sandbox` tags in the merge confirmation modal.
- Three typo fixes (`seperate`, `worklow`, `pushed`).

## New sections

- **Who can do what** role table reflecting the visibility/access rules from [OpenFn/lightning#4763](https://github.com/OpenFn/lightning/pull/4763): direct membership is required to see or open a sandbox; merge requires admin/owner on the source plus editor+ on the target; `User.role` is a user type, not a project role.
- **Restoring a deleted sandbox** covers the soft-delete + grace-period UX.
- **Limits** subsection: active sandbox count cap, nesting depth cap.
- **Per-tab Settings inheritance** breakdown matching the four banner variants (`:local`, `:editable`, `:inherited`, plus the parent-managed Webhook Security tab).
- **Webhook authentication methods** clarified as parent-owned and parent-managed.
- **Channels are not cloned** at provision (explicit note).
- **CLI: multi-sandbox tips** — `openfn project fetch` auto-aliases sandboxes; `openfn project checkout` switches between them.

## AI Usage

- [x] I have used Claude Code
- [ ] I have not used AI


docs/build/sandboxes.md
```diff
@@ -7,177 +7,306 @@ Sandboxes are a way to develop fixes and new features on workflows without
...
+| See a sandbox                | Direct `project_users` row, or support user with `allow_support_access` on the sandbox |
+| Open a sandbox               | Same as above                                                                          |
+| Create a sandbox             | `editor`, `admin`, or `owner` on the parent                                            |
+| Edit or delete a sandbox     | `admin` or `owner` on the sandbox itself, or `admin`/`owner` on the workspace root     |
+| Merge a sandbox              | `admin` or `owner` on the source, and `editor`+ on the merge target                    |
+| Cancel a scheduled deletion  | Same as edit or delete                                                                 |
+
+A few things follow from these rules:
+
+- A user with a role on a parent project will not automatically see sandboxes
```
josephjclark: Eh this is half true. When we create a sandbox we also copy the relationships. This is true on staging.

So adding a new relationship to the parent won't update existing sandboxes, correct. But  existing users on a parent are automatically added to the sandbox.

docs/build/sandboxes.md
```diff
@@ -7,177 +7,306 @@ Sandboxes are a way to develop fixes and new features on workflows without
...
+  under it. They need to be explicitly added to each sandbox they should
+  access.
+- When you can see a sandbox but not its parent, the breadcrumb at the top of
```
josephjclark: I wouldn't even mention this tbh. Too detailed and I don't see the value to users?

docs/build/sandboxes.md
```diff
@@ -7,177 +7,306 @@ Sandboxes are a way to develop fixes and new features on workflows without
...
+  the app starts at the highest project you can see (rather than the actual
+  workspace root). Ancestors above that point are silently hidden.
+- The global project picker hides projects you can't see. When you can see a
```
josephjclark: I would either remove this entirely or stop at `The global project picker hides projects you can't see`

docs/build/sandboxes.md
```diff
@@ -7,177 +7,306 @@ Sandboxes are a way to develop fixes and new features on workflows without
...
+  sandbox but not the sandboxes between it and the root, the picker collapses
+  the gap and nests the sandbox under the nearest visible ancestor.
 
 ## Viewing a sandbox
 
-To develop and test a sandbox, you need to enter it from the Sandboxes item on
-the main navigation bar.
+To open a sandbox, click your project name in the top-left breadcrumb to open
```
josephjclark: This is only true if you're in a project already.

it's a bit complex to explain. I might just say:
```
You can see all sandboxes on a project from the left-hand menu. From here, click a sandbox name to enter it. You can also enter a sandbox from the project picker (ctrl/CMD + P).

When you're inside a sandbox....
```

docs/build/sandboxes.md
```diff
@@ -7,177 +7,306 @@ Sandboxes are a way to develop fixes and new features on workflows without
...
+the global picker, then pick the sandbox. You can also reach the sandbox from
+the **Sandboxes** page of any project where you can see it.
+
+When you're inside a sandbox, the app will change color to help you understand
```
josephjclark: The app doesn't change colour, only the picker does 💔

docs/build/sandboxes.md
```diff
@@ -7,177 +7,306 @@ Sandboxes are a way to develop fixes and new features on workflows without
...
 
 When merging, we replace the contents of workflows in the target project with
-those in your sandbox. Renaming a worklow will make it look like the workflow
+those in your sandbox. Renaming a workflow will make it look like the workflow
 was removed from the base, and a new workflow added.
 
-After merging, the sandbox will be destroyed, along with its history,
-collections and dataclips. Any environments and credentials associated with the
-project will be unaffected.
+Collections are synced by name. Collections with the same name on the source
```
josephjclark: I find `collections are synced by name` to be confusing. Because collections don't sync.

I would be more vague here.

Maybe `Collections will merge relationships but not data. So if you add a new collection to a sandbox, the target project will automatically have a new, empty collection created. Similarly if a collection is removed from a sandbox, it will be removed from the target project on merge.`

docs/build/sandboxes.md
```diff
@@ -7,177 +7,306 @@ Sandboxes are a way to develop fixes and new features on workflows without
...
+and target are left alone (their data isn't copied). Collections that exist
+only on the source are created empty on the target. Collections that exist
+only on the target are deleted along with their items.
+
+To merge a sandbox you need to be `admin` or `owner` of the **source**
```
josephjclark: "source" isn't clear - maybe `source (the sandbox you are merging from)`

josephjclark (commented):
Thanks @elias-ba, some great catches here. I've left a couple of suggestions

docs/build/sandboxes.md
elias-ba: Fixed. You're right,  the parent's `project_users` are copied across at provision time, with parent owners demoted to admin and the creator becoming the sandbox owner. The accurate behaviour is already covered in the 'Creating sandboxes' section, so I dropped the misleading bullet here.

docs/build/sandboxes.md
elias-ba: Agreed. Dropped this bullet entirely.

docs/build/sandboxes.md
elias-ba: Trimmed to just `The global project picker hides projects you can't see.`

docs/build/sandboxes.md
elias-ba: Rewrote along the lines you suggested. Now leads with the project's Sandboxes page in the side menu, and mentions Ctrl/Cmd+P.

docs/build/sandboxes.md
elias-ba: You're right. Changed

docs/build/sandboxes.md
elias-ba: Rewrote to "Collections are merged by relationship, not by data". Much clearer, agreed. I kept one short line about same-name collections being left alone, because that has a real consequence for users (their items on each side stay where they are, neither side overwrites the other). Open to dropping it if you'd rather keep the paragraph minimal.

docs/build/sandboxes.md
elias-ba: Done. Now reads **source** (the sandbox you're merging from).

elias-ba:
Thanks for the careful review, I pushed a commit addressing all seven inline comments. While I was in there, I also did an audit pass on the rest of the doc and caught a few more mismatches that I corrected. Can you give me another round of review ? Feel free to edit and push commits too.

josephjclark (approved):
Thanks @elias-ba

PR #800, opened 2026-06-11
security updates and fixed particles

josephjclark:
Bumps various dependencies with `yarn up \*`. Replaces most of the open dependabots (I think they'll close themselves?)

One thing to note is that `particles` (which runs on the homepage) wasn't a straightforward update. It looks a bit worse on this branch?

Probably need good QA on this branch tbh


josephjclark:
Ok, so particles is being an absolute pain
- it doesn't link to its own correct doc for the react plugin - so it's hard to find the right docs
- The docsite is full of popup ads which makes me _so angry_
- When switching page the component doesn't get removed, so you see the particules drift over over docs pages. Super annoying. I've tried a quick fix but it's not working.

I don't think there are any security alerts related to particles or the react version so I'm just gonna shut this down. Frustrating.

PR #841, opened 2026-08-25
Add unit testing guide

### LINKED ISSUE #840: add a guide on unit testing job code
## Context

`openfn compile` now writes compiled job expressions to disk, which means job code can finally be imported into a normal JavaScript test runner. Job expressions are not valid JS out of the box (top-level adaptor calls like `get('/endpoint')` break
the import), so until now there was no first-class way to unit test the helper functions users write inside their jobs.

There's an internal guide in kit covering the mechanics:
https://github.com/OpenFn/kit/blob/main/.claude/unit-testing-jobs.md

That guide is written for engineers who already know the compiler. We need a user-facing version on docs.openfn.org that is task-oriented: "here is how you test your job code", not "here is what `--exports-only` does".

## Problem

Users writing non-trivial transformation logic (date formatting, SMS parsing, identifier mapping, code lookups) currently have no documented way to test it except by running the whole workflow with the CLI and eyeballing `output.json`. That's slow, doesn't isolate failures, and doesn't run in CI.

## Proposed change

Add a new page: **Writing unit tests for your jobs**

Suggested location: alongside the job-writing guide, e.g. `documentation/jobs/unit-testing-jobs`, so it sits with the other "how to write good job code" content rather than buried in the CLI reference.

Cross-links needed from:

- `documentation/jobs/job-writing-guide` — a short "your helpers can be unit tested" pointer
- `documentation/cli-usage` — add `compile` to the common-usage examples with a link here
- `documentation/cli` — mention testing in the "you can use the CLI to…" list

## Content outline

1. **Why unit test job code** — what is and isn't testable. Operations
   (`fn`, `get`, `each`) are not unit-testable; pure helper functions are. Set
   expectations up front.
2. **Write your helpers so they're testable** — the most important section, and
   the one missing from the kit guide. In `--exports-only` mode only
   `export const` and `export function` declarations survive; a plain
   `const formatDate = ...` is dropped. So the guidance is: **export any helper
   you want to test.** Include a before/after snippet.
3. **Compile your workflows** — `openfn compile --exports-only`, what lands in
   `dist/`, the `.mjs` extension (and why that means you don't need
   `"type": "module"`), and `-o` / `--clean` / `--workspace`.
4. **Write a test** — a full worked example, source → compiled → test, using
   Node's built-in test runner (`node --test`). Reuse the `formatDate` example
   from the kit guide or write a slightly richer one (e.g. parsing an SMS string
   into a structured record) so it demonstrates real value.
5. **The edit → test loop** — `--watch` alongside the test runner's watch mode.
6. **Wire it into your project** — `package.json` scripts, `.gitignore` the
   output dir, and a short note on running this in CI (GitHub Actions snippet
   would be a nice-to-have, not blocking).
7. **Reference table** — `compile` flags: `--exports-only`, `-o`, `-O`,
   `--watch`, `--clean`, `--workspace`, plus the `dirs.compiled` key in
   `openfn.yaml`.
8. **Troubleshooting** — see below; this is where most of the new writing is.

## Gotchas that must be covered

These are the things that will generate support questions if we skip them:

- **"My helper isn't in the compiled file."** It wasn't exported. Non-exported
  declarations are always dropped in strip mode.
- **"No file was written for my step."** Steps whose compiled output is empty
  after stripping are skipped entirely, so the import fails with a
  module-not-found error rather than anything descriptive.
- **Adaptor imports are preserved in compiled output.** A step that does
  `import { dateFns } from '@openfn/language-dhis2'` will still carry that import
  after compilation, so the adaptor package has to be installed locally for the
  test to run. This needs an explicit "install your adaptors as devDependencies"
  instruction, or a documented pattern for keeping testable helpers free of
  adaptor imports. **This is the biggest open gap in the current guide.**
- **`export default` is removed** in strip mode — expected, since it only exists
  for the runtime.
- **Import paths in tests** point at the compiled output (`../dist/...`), not the
  source. Worth stating plainly, and worth noting that the compiled dir should be
  gitignored so tests depend on a build step.
- Full compilation (no `--exports-only`) writes every step and keeps operations
  in `export default [...]`. Mention it briefly so users know the flag matters,
  but don't make it the focus.

## Acceptance criteria

- [ ] New page published under `documentation/` and added to the sidebar
- [ ] Page includes a complete, copy-pasteable worked example (source → compiled → test → command to run it)
- [ ] All CLI commands and flags in the page have been run against the current released `@openfn/cli` and produce the documented output
- [ ] The adaptor-dependency gotcha is documented with a working solution, not just a warning
- [ ] All eight gotchas above are covered somewhere on the page
- [ ] Cross-links added from the job-writing guide and CLI usage pages
- [ ] Minimum required `@openfn/cli` version stated on the page

## Out of scope

- Integration/end-to-end testing of whole workflows (running with fixture state,
  asserting on final state). Worth a follow-up page; keep this one focused on
  unit tests.
- Mocking adaptor operations or HTTP calls.
- Recommending a specific third-party test runner. Use `node --test` in examples
  and note that any runner works.

## Open questions

- Which `@openfn/cli` version first shipped `compile` with `--exports-only`? Needs
  confirming against the kit release notes / changelog before the page states a
  minimum version — I don't want to guess at that.
- Does the compiled output dir get a `.gitignore` automatically (as `.cli-cache`
  does), or does the user need to add it? Affects the "wire it into your project"
  section.
- Should `compile` also get a proper entry in the CLI reference docs as part of
  this issue, or as a separate one?
- Is there an existing testing/quality section in the docs IA this should live
  under, rather than a standalone page?

## Source material

- https://github.com/OpenFn/kit/blob/main/.claude/unit-testing-jobs.md

mtuchi:
## Short Description

Adds a new docs page, "Writing unit tests for your jobs", covering how to compile job code with `openfn compile --exports-only` and test the helper functions with a standard JavaScript test runner. Cross-linked from the job writing guide and the two CLI pages.

Closes #840 

## Details

- New page `docs/jobs/unit-testing-jobs.md`, added to the sidebar under **Write Jobs**, after `jobs/compilation`
- `docs/jobs/job-writing-guide.md` picks up a short "your helpers can be unit tested" pointer in Next Steps
- `docs/build-for-developers/cli-intro.md` gains a bullet in the "you can use the OpenFn CLI to..." list
- `docs/build-for-developers/cli-usage.md` gains a "Compile job code for unit testing" section

[Walk through video](https://docs.google.com/videos/u/0/d/1NuIKwGme0ksOWPlZ16aDUWK8wdkzEYg5JwuPElKg-5o/play)

## AI Usage

- [x] I have used Claude Code
- [ ] I have not used AI

[Responsible AI Policy](https://www.openfn.org/ai)


docs/jobs/job-writing-guide.md
```diff
@@ -55,6 +55,12 @@ If you're ready to start using the app, take a look at this guide to
 Workflow design is a non-trivial problem, so you might also like to review the
 Workflow [Design Process docs](/documentation/design/design-overview).
 
+As your jobs grow, you'll start writing helper functions inside them - parsing,
+mapping, reformatting. Those helpers can be unit tested like any other
+JavaScript: export them, compile your project with the CLI, and point a test
```
PiusKariuki: Might want to let the know that the export keyword can only be used for functions on the top level of the job code.

PiusKariuki (commented):
This is so good @mtuchi . I have just one minor comment for you.

docs/build-for-developers/cli-usage.md
```diff
@@ -222,6 +225,49 @@ running workflows via the CLI.
...
+a test runner. `openfn compile` writes them out as ordinary ES modules.
+
+**Compile every workflow in the project, keeping only exported declarations:**
+
+```bash
+openfn compile --exports-only
+```
+
+Compiled files are written to `dist/` as `.mjs`, mirroring your workflow
+folders. Operations (`fn`, `get`, `each`) are stripped, so what's left is the
```
josephjclark: This bit about operations is SUPER important

I think we need some kind of introduction saying that you can unit test the pure JS functions of your job code, but not the operations themselves.

This isn't just a detail or a caveat, it's core to understanding what unit testing a job means

docs/build-for-developers/cli-intro.md
```diff
@@ -14,6 +14,9 @@ developer experience with OpenFn. You can use the OpenFn CLI to:
 - Sync workflows between OpenFn and a local filesystem or GitHub
 - Securely run OpenFn workflows
 - Troubleshoot and debug OpenFn steps
+- Compile your job code so you can
```
josephjclark: You've build the docs around compilation - but users don't know what compilation is.

This needs to just be "Unit testing job code". And compilation needs to be a detail "You can unit test pure functions in your job code by compiling them to native js, and importing them into a test file" or something like that

docs/build-for-developers/cli-intro.md
```diff
@@ -14,6 +14,9 @@ developer experience with OpenFn. You can use the OpenFn CLI to:
...
+  [unit test it](/documentation/jobs/unit-testing-jobs) in a standard JavaScript
+  test runner
```
josephjclark: remove `test runner`, it;s just part of the unit testing story

docs/build-for-developers/cli-usage.md
```diff
@@ -222,6 +225,49 @@ running workflows via the CLI.
 
 ---
 
+### Compile job code for unit testing
+
+Job expressions aren't valid JavaScript, so you can't import them straight into
```
josephjclark: I've alluded to this a few times but this sort of starts in the middle of the problem I think. Too technical. Start by outlining the user story "So you want to write unit tests against your job code? Start here..."

docs/build-for-developers/cli-usage.md
josephjclark: Also you can compile without `exports-only` and  the operations are left in your code. They're just unusable!

docs/build-for-developers/cli-usage.md
```diff
@@ -222,6 +225,49 @@ running workflows via the CLI.
...
+openfn compile my-workflow --exports-only
+```
+
+**Print the compiled output instead of writing files:**
+
+```bash
+openfn compile path/to/job.js -a http -O
+```
+
+**Recompile whenever a source file changes:**
```
josephjclark: What's a source file? You mean step code?

docs/jobs/job-writing-guide.md
josephjclark: Maybe (although this is a basic Javscript rule so maybe not), but not right here. This is just a high level mention of what you can do, nto a step-by-step guide to how.

But I think you're right: we don't really explain that if you want to test a function, you probably ought to declare it as an export.

docs/jobs/unit-testing-jobs.md
```diff
@@ -0,0 +1,529 @@
+---
+sidebar_label: Unit Testing Jobs
+title: Writing unit tests for your jobs
+---
+
+Most job code goes like this: fetch some records, reshape them, send them
+somewhere else. But the reshaping bit often grows into real logic - parsing an
```
josephjclark: I'm being picky but I'd rephrase "real logic" to "complex logic"

josephjclark (changes requested):

docs/build-for-developers/cli-usage.md
mtuchi: @josephjclark, there are some weird hack you have to do if your job code is using collections
i was trying to avoid that

docs/jobs/unit-testing-jobs.md
mtuchi: thanks

docs/jobs/job-writing-guide.md
mtuchi: i have update to say `export them at top level,...

docs/build-for-developers/cli-usage.md
mtuchi: job code

docs/build-for-developers/cli-usage.md
mtuchi: I have updated the section

docs/build-for-developers/cli-intro.md
mtuchi: I have update to use `Unit testing job code` instead. but  i reused the `You can unit test pure functions in your job code by compiling them to native js, and importing them into a test file` some place else

docs/build-for-developers/cli-usage.md
mtuchi: I have break down what it means "job code unit testing" and explain what is testable

docs/build-for-developers/cli-usage.md
```diff
@@ -222,6 +225,70 @@ running workflows via the CLI.
 
 ---
 
+### Compile job code for unit testing
+
+So you want to write unit tests against your job code? Start here.
```
josephjclark: Should you start here? Or should you start in `unit-testing-jobs`?

I've just looked closely at where these docs are and they're fartoo verbose for the basic usage section of the CLI.

You should just list the basic commands for testing and then refernce the main `unit-testing-jobs` page

docs/jobs/unit-testing-jobs.md
```diff
@@ -0,0 +1,528 @@
...
+openfn project pull <uuid>
+```
+
+That gives you a folder with `openfn.yaml`, a `workflows/` directory, and one
+`.js` file per step. See [OpenFn Sync](/documentation/sync) for pulling,
+checking out and deploying projects.
+
+:::
+
+## Why you need to compile first
```
josephjclark: We haven't mentioned compilation yet. Why are you talking about compilation right at the top of the unit testing docs? Confusing for new readers.

docs/jobs/unit-testing-jobs.md
```diff
@@ -0,0 +1,528 @@
...
+```
+
+There is no `import` for `fn`, and the bare `fn(...)` call at the top level
+would execute on import even if there were. See
+[Compilation](/documentation/jobs/compilation) for the full explanation.
+
+`openfn compile` turns job expressions into ordinary ES modules on disk. Once
+that's done, your test runner can import them like any other JavaScript.
+
+## What you can and can't unit test
```
josephjclark: I get the point but this feels a little over-explained? The table makes this feel way more complicated than this is

docs/jobs/unit-testing-jobs.md
```diff
@@ -0,0 +1,528 @@
...
+Operations are stripped out during compilation, so there is nothing left to
+import. To exercise a whole step or workflow, run it with
+`openfn <workflow-name> -s tmp/input.json` and inspect the output state.
+
+The trick, then, is to move your logic out of operations and into functions, and
+then test the functions.
+
+## Step 1: Export any helper you want to test
+
+This is the most important thing on this page.
```
josephjclark: Is it? 🤔

docs/jobs/unit-testing-jobs.md
```diff
@@ -0,0 +1,528 @@
...
+
+When you compile with `--exports-only`, the compiler keeps **only exported
```
josephjclark: We haven't explained compuling with `--exports-only` yet, so it's weird that we're explaining this now

docs/jobs/unit-testing-jobs.md
```diff
@@ -0,0 +1,528 @@
...
+declarations**. Everything else - operations, plain `const`s, plain
+`function`s - is dropped.
+
+So a helper you want to test must be exported:
+
+```js title="Before - not testable"
```
josephjclark: I would ditch this Before example.

docs/jobs/unit-testing-jobs.md
```diff
@@ -0,0 +1,528 @@
...
+
+fn(state => ({ ...state, data: state.data.messages.map(parseSms) }));
+```
+
+The job runs identically either way - `export` is a no-op at runtime. It only
+changes what survives compilation.
+
+:::warning Export the things your helpers depend on, too
+
+Stripping is per-declaration, not "everything `parseSms` touches". If you export
```
josephjclark: What is stripping?

docs/jobs/unit-testing-jobs.md
```diff
@@ -0,0 +1,528 @@
...
+    upload.js
+dist/
+  sms-intake/
+    parse-message.mjs     # compiled
+```
+
+Two things to note:
+
+- **Output files use the `.mjs` extension.** Node always treats `.mjs` as an ES
+  module, so you don't need `"type": "module"` in your `package.json` for the
```
josephjclark, line 59: Wait why do I suddenly need a package json? That wasn't mentioned in requirements

docs/jobs/unit-testing-jobs.md
```diff
@@ -0,0 +1,528 @@
...
+
+```yaml title="openfn.yaml"
+dirs:
+  workflows: workflows
+  compiled: workflows
+```
+
+## Step 3: Write a test
+
+Here's the whole loop, end to end.
```
josephjclark: Loop of what?

docs/jobs/unit-testing-jobs.md
```diff
@@ -0,0 +1,528 @@
...
+
+### The source
```
josephjclark: Why are we duplicating the source and output? Didn't we just explain all of this?

docs/jobs/unit-testing-jobs.md
```diff
@@ -0,0 +1,528 @@
...
+export const parseSms = text => {
+  const parts = text.trim().split('#');
+  return FIELDS.reduce((record, field, i) => {
+    record[field] = parts[i]?.trim() ?? null;
+    return record;
+  }, {});
+};
+```
+
+The `fn(...)` operation is gone. Both exports survived.
```
josephjclark: Reads a lot like AI generated test and makes me a bit uncomfortable. It's the short sentences and the braggy justification of "Both exports survived"

docs/jobs/unit-testing-jobs.md
```diff
@@ -0,0 +1,528 @@
...
+
+### The test
+
+Note the import path: it points at `dist/`, **not** at your source file.
```
josephjclark: Shouldn't this be a comment inline, next to the code it  describes?

docs/jobs/unit-testing-jobs.md
```diff
@@ -0,0 +1,528 @@
...
+
+```js title="test/parse-message.test.mjs"
+import { test } from 'node:test';
```
josephjclark: Do I have to use `node:test`?

We don't need to explain unit test frameworks but I think it's worth a mention that users can use any runner they want . We're just using node as an example

docs/jobs/unit-testing-jobs.md
```diff
@@ -0,0 +1,528 @@
...
+import assert from 'node:assert/strict';
+
+import { parseSms } from '../dist/sms-intake/parse-message.mjs';
+
+test('parses a well-formed message into a record', () => {
```
josephjclark: We only want one test for this example

docs/jobs/unit-testing-jobs.md
```diff
@@ -0,0 +1,528 @@
...
+  assert.equal(record.id, 'P-002');
+  assert.equal(record.name, 'Grace Hopper');
+});
+
+test('fills missing trailing fields with null', () => {
+  assert.equal(parseSms('P-003#Alan Turing').dob, null);
+});
+```
+
+### Run it
```
josephjclark: Run what? The test or the step?

docs/jobs/unit-testing-jobs.md
```diff
@@ -0,0 +1,528 @@
...
+
+```bash
+openfn compile --exports-only && node --test
```
josephjclark, line 173: Why is this two commands? We've already talked about the compile step.

The run command also only applies to our nodejs test runner. That's a bit of a problem with this guide. I suppose users will know how to adapt to what they're using?

docs/jobs/unit-testing-jobs.md
```diff
@@ -0,0 +1,528 @@
...
+```
+✔ parses a well-formed message into a record (0.9ms)
+✔ trims whitespace around each field (0.1ms)
+✔ fills missing trailing fields with null (0.1ms)
+ℹ tests 3
+ℹ pass 3
+ℹ fail 0
+```
+
+We use Node's built-in test runner here because it needs no dependencies, but
```
josephjclark: Right, we have this. This needs to go way earlier in the guide

docs/jobs/unit-testing-jobs.md
```diff
@@ -0,0 +1,528 @@
...
+nothing about this is Node-specific. Vitest, Jest, Mocha, `node:test` - any
+runner that can import an ES module will work.
+
+:::tip Name your test files `.test.mjs`
```
josephjclark: Is any of this relevant? What if I'm writing tests in typescript?

docs/jobs/unit-testing-jobs.md
```diff
@@ -0,0 +1,528 @@
...
+
+The compiled output is `.mjs` and needs no configuration. Your _test_ files are
+yours, though - if you name them `.js` in a project without `"type": "module"`,
+Node will warn about reparsing them as ES modules. Naming them `.test.mjs`
+avoids the warning without touching your `package.json`.
+
+:::
+
+## Step 4: The edit → test loop
```
josephjclark: Irrellevant, no?

docs/jobs/unit-testing-jobs.md
```diff
@@ -0,0 +1,528 @@
...
+And your test runner in watch mode in another:
+
+```bash
+node --test --watch
+```
+
+Now editing a step recompiles it, which changes a file in `dist/`, which re-runs
+your tests.
+
+## Step 5: Wire it into your project
```
josephjclark: Also irrellevant, no?

docs/jobs/unit-testing-jobs.md
```diff
@@ -0,0 +1,528 @@
...
+
+Add scripts to your project's `package.json`:
+
+```json title="package.json"
+{
+  "devDependencies": {
+    "@openfn/cli": "^1.39.0"
```
josephjclark: Why is the CLI a dev dependency of the package? We usually tell users to just install it globally

docs/jobs/unit-testing-jobs.md
```diff
@@ -0,0 +1,528 @@
...
+### Gitignore the compiled output
+
+**The CLI does not create a `.gitignore` for the compiled directory.** (It does
+for `.cli-cache`, which is why you may expect otherwise.) Add it yourself:
+
+```title=".gitignore"
+dist/
+```
+
+The generated `.mjs` files are build output: every one of them can be recreated
```
josephjclark: We've explained this right? Drop it

docs/jobs/unit-testing-jobs.md
```diff
@@ -0,0 +1,528 @@
...
+from your `.js` steps with a single `openfn compile --exports-only`. Committing
+them means a second copy of every helper in the repo, a diff on every step edit,
+and merge conflicts in files nobody wrote by hand - plus the risk that a stale
+`dist/` gets reviewed as if it were source.
+
+If you compile to somewhere other than `dist/` - with `-o build`, or a different
+`dirs.compiled` in `openfn.yaml` - ignore that folder instead. Your `workflows/`
+folder and `openfn.yaml` are source, and should stay tracked.
+
+### Install your adaptors as devDependencies
```
josephjclark: hmm. Isn't this only here if you don't do `--exports-only`?  This feels like a real stretch. Also the example  assumes github I think. What if you're using circleCI? GitLab? Azure something?

docs/jobs/unit-testing-jobs.md
```diff
@@ -0,0 +1,528 @@
...
+      - uses: actions/setup-node@v4
+        with:
+          node-version: 24
+      - run: npm ci
+      - run: npm test
+```
+
+`npm test` runs `compile` first, so the tests always run against fresh output.
+
+## Troubleshooting
```
josephjclark: I don't love it - are you sure all these are useful problems to include all these docs for? It  feels like bloat to me.

docs/jobs/unit-testing-jobs.md
```diff
@@ -0,0 +1,528 @@
...
+export default [post('/patients', { body: state.data })];
+```
+
+That's the code the runtime actually executes, and it's useful for understanding
+or debugging compilation. It is not useful for unit testing: importing it pulls
+in the adaptor and evaluates the operations. Use `--exports-only` for tests.
+
+## Reference
+
+### `openfn compile` flags
```
josephjclark: Maybe better to just say `run openfn compile --help` here? No way will I remember to update this list when updating the compile command

docs/jobs/unit-testing-jobs.md
```diff
@@ -0,0 +1,528 @@
...
+
+### `openfn.yaml` keys
+
+```yaml
+dirs:
+  workflows: workflows # where the CLI looks for workflows
+  compiled: dist # where openfn compile writes output
+```
+
+## Next steps
```
josephjclark: None of these are "next steps". They might be Related  Pages I guess?

docs/jobs/unit-testing-jobs.md
```diff
@@ -0,0 +1,528 @@
...
+
+- [Compilation](/documentation/jobs/compilation) - what the compiler does and
+  why
+- [Best Practices](/documentation/jobs/best-practices) - writing job code that's
```
josephjclark, line 207: That's a great point  does our best practice doc encourage writing testable functions?

josephjclark (commented):
AI isn't ready to write out docsite yet. 

I don't like it. The more closely I read this the more problems I find :(

I think I'd rather have a really short, concise guide to unit testing which doesn't have all this explanation. It's long and repetetive and confusing - I'm not convinced we need 500 lines of this.

docs/jobs/unit-testing-jobs.md
mtuchi: Updated the best practices to include `Write testable functions`

docs/jobs/unit-testing-jobs.md
mtuchi: I have removed the Troubleshooting guide, will comeback to it when there is a need

mtuchi: Running in CI is important

docs/jobs/unit-testing-jobs.md
mtuchi: No, so if you have a function that uses adaptor functions then you have to install that adaptor. For example
```
export const formatDate = (date) => dateFns.format(date, 'yyyy-MM-dd');
```

dateFns is a function from adaptor so you have to install that adaptor

docs/jobs/unit-testing-jobs.md
mtuchi: You have to compile every time you make changes to job file running these commands make sure you always have the latest changes from the workflow. 

The altenative is running one the compile command in watch mode and in another terminal run `node --test`

docs/jobs/unit-testing-jobs.md
mtuchi: No it's important

docs/build-for-developers/cli-usage.md
mtuchi: I have updated the section

docs/jobs/unit-testing-jobs.md
mtuchi: i have updated the section

mtuchi:
Hiya @josephjclark i have cut down lots of verbose stuff, please have another look

